PR #41, opened 2023-03-21
Move intent

hypnoshock:
# What

Clicking the move button on a seeker will set the intent on the game state to `move` by calling `setIntent` on the `core-client`. The map reads this intent on the next state update which puts the map into a `isMoving` context. When in the `isMoving` context, tiles selected for the movement path are appended to the `selectedTiles` array in the game state. Unlike before, Unity does not highlight tiles as a direct side effect of clicking on the tile but instead highlights tiles according to the state that is handled in in the `OnStateUpdated` listener.

- The core-client state object now has a string field for intent which is interpreted by the UI
- At any time `cancelIntention()` can be called set the intention back to `NONE` and the intentTileIDs to an empty array

# Why

The map component held all the state related to movement paths which meant that any external plugins couldn't see that state. Also future plugins should be able to specify a path and the map display it

# Other bits

- The switch which handles the messags within the bridge and in the Unity organism have been made the same. At some point I would like to make that into a class that can be shared between the two modules so I don't have to keep two things up to date. At least they work the same now and can be copied and pasted at least
- I updated `StateSchema.cs` by rebasing `state-schema-codegen` onto this branch, running the codegen, stashing, switching back to this branch and committing. I think the `state-schema-codegen` tool could be merged into main without any danger 
- Map manager no longer checks a `stateUpdated` flag in the update but instead reacts directly to the `EventStateUpdated` event that comes from the `PluginManager`
  - It was there because the event from the PluginManager used to come from another thread but now the mechanism of sending the event in the main thread is handled by the `PluginManager` so no need to accommodate for that in any of the classes that subscribe to this event 

# Known issues
- Right clicking on the final tile to close the path employs a hack to add the final tile directly to the `_path` array within the `SeekeMovementManager` instead of adding it to `selectedTiles` and waiting for the state to be updated. Closing the path clears the player's selection anyway so right now not a problem.
- The execution of movement is still happening via the co-routine instead of submitted as a MOVE_MULTI action. 
- The `intent` field perhaps should be hoised up a level in the state object as it doesn't belong in the selected object

### DIFF
--- a/core/src/client.ts
+++ b/core/src/client.ts
@@ -35,16 +35,27 @@ const ACTIONS = new ethers.Interface([
 
 const gameID = 'DAWNSEEKERS';
 
+export enum Intention {
+    NONE,
+    MOVE,
+    SCOUT,
+    BUILD,
+}
+
 export interface RawSelectionState {
     playerAddr?: string;
     seekerID?: NodeID;
     tileIDs: NodeID[];
+    intention: Intention;
+    intentTileIDs: NodeID[];
 }
 
 export interface SelectionState {
     player?: Player;
     seeker?: Seeker;
     tiles?: Tile[];
+    intention: Intention;
+    intentTiles?: Tile[];
 }
 
 export interface Session {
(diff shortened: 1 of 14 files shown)


core/src/client.ts
```diff
@@ -323,6 +331,17 @@ export class Client {
         return this.publish();
     }
 
+    async cancelIntent() {
```
chrisfarms, line 334: I don't think we should have an explicit "cancelIntent" command... since we can do the same thing with `setIntent(NONE)` ... which is the same way selectSeeker(undefined) and selectedTiles(undefined) work

I also don't think it should clear the tilesID selection ... because:

* if two things are switching between intents you don't want it to loose your other selections.
* it couples "intent" too closely to tiles ... it could be an intent for a seeker (like "attack")

hypnoshock: Removed from Client and the two places that referenced it 👍

JackPlaymint (approved):

PR #42, opened 2023-03-22
@dawnseekers/core: stream all the things

chrisfarms:
### What

if you are looking for an entrypoint then [the react context helper](https://github.com/playmint/ds/pull/42/files#diff-d0beedec9b327156a0a96a3468b7eaf9f50de779b1f69d01e2b10ea9b8722ea1R56-R72) is probably a good starting point as it shows how everything is wired together.

Core:

* Stream All The Things: this is a rewrite of the prototype/core lib to be based on reactive/observable/streams to allow managing the flow of dependent async things ... we are using [Wonka](https://wonka.kitten.sh/api/) to help with this for this, chosen because (1) it's already a dep we have, and (2) it has a tiny surface area (unlike RxJS or bigger state management libs)
* Generally the idea of their being one giant "client" is gone ... it's more like a construction kit for the bits of the client you need ... the react helpers make it look bit like one giant thing tho so in reality you don't really see the "streams" in React land (unless you want to)
* Subscriptions to trigger state change replace the hacky polling (requires https://github.com/playmint/cog/pull/6)
* State split into two parts; "world" and "selected" (world is a shallow version of everything, selected things are deeper version of currently selected things)
* Plugins, (follow similar pattern to state/selected split) AvailablePlugins are discovered in the background and autoloaded if relevent. there's a selectPlugin call to active plugins by id ... you can hard code plugins on startup (example for scout), but this is just for testing - all plugins should come from chain discovery
* Wallet handling improved and anything that requires it should switch along with the streams on changes
* Dispatch queues ... the dispatch() is now tied to the currently connected Player, and gets it's own queue to ensure dispatches go in order (and allow subscribing to the queue events)
* Selection is implemented differently see `makeSelector` ... but the pattern is similar ... every selection thing has a pair of `source, selectThing()` ... I have added `selectIntent` as an example of the simplest selector
* Logs ... there is now a Logger implementation that can send logs to a stream for later subscription ... plugins get passed a sub-logger of this logger so they can send messages to the screen .... the dispatcher is currently wired up with a logger too .... this all needs work, but it's a start
* New React helpers .... there's a React context provider + a bunch of usePlayer, useWorld, useLogs hooks for subscribing to state .... the `useSource` hook is a generic helper for subscribing to any of the available sources ... **the context provider in react.ts is probably the best place to start to see how all the streams are wired together.**
* the package builds as a commonjs/esm module ... so it is publishable
* probably more I've forgotten...

Frontend:

* fixups for the new shape of the state
* hack to map the newState to oldState shapes (ish - it's lossy so map will need fixing properly)
* adds basic tile info component for constructing + viewing whats on the tile 
* show type=building autoloaded plugins in the tile info component
* use the new "intents" for turning on/off the constuction panel (so that it can be done from the map controls soon)

Bridge:

* rejigged for new core bits

### Still working on

* [x] fixing up construct button
* [x] trying to write some tests and add more docstrings
* [x] rebase everyone's stuff and fix what I broke
* [x] make things prettier 
* [x] remove all the console.log noise


core/src/react.tsx
```diff
@@ -0,0 +1,180 @@
...
+        const { plugins: enabledPlugins, selectPlugins } = makePluginSelector(client, defaultPlugins);
+        const { plugins: autoloadPlugins } = makeAutoloadPlugins(client, availablePlugins, selection);
+
+        const uiPlugins: Source<PluginConfig[]> = pipe(
+            zip([enabledPlugins, autoloadPlugins]),
+            map((plugins) => [...plugins[0], ...plugins[1]]),
+        );
+        const ui = makePluginUI(player, logger, uiPlugins, world, selection);
+
+        console.log('USE MEMO');
```
chrisfarms, line 73: remove log

core/src/plugins.ts
```diff
@@ -0,0 +1,506 @@
...
+                }
+
+                // stick the refs somewhere we can find them later
+                globalThis.__refs = refs;
+
+                return JSON.stringify(res);
+
+            })(${JSON.stringify(state)})`);
+        const pluginResponse = JSON.parse(context.unwrapResult(res).consume(context.getString));
+        console.debug(`[${pluginId} recv]`, pluginResponse);
```
chrisfarms, line 464: rm debug

hypnoshock (approved):
Excellent work with the streams. I don't have anything to add to this

PR #43, opened 2023-03-24
State schema codegen

hypnoshock:
# What

A simple cli c# tool `state-schema-gen` which takes a json-schema and outputs C#
The json-schema that the tool consumes can be generated using `typescript-json-schema` from within the core folder.
The process of how to generate the json schema and turn that into c# class is listed in detail below.
It instructions assume that the interface that we are referencing is named `MapState` which doesn't exist yet. Previously I had aliased the the `State` interface in `client.ts` but this won't exist after the soon to come big state change merge.

(These instructions are also in `README.md` in the `state-schema-gen`)

# Codegenning the schema

Make sure `typescript-json-schema` is installed in `core`
`npm install typescript-json-schema -g`
It wasn't adding as dev dependency when I tried to do this and I'm not sure why

Build core
I couldn't do this in isolation and had to build the whole game

Run the json schema generator from the `core` folder
`npx typescript-json-schema tsconfig.json MapState -o dist/json-schema.json`

NOTE: If it failes due to bigint not being serialisable, You need to add the following decorator to the field
`/** @TJS-type string */`

Build the release version of `state-schema-gen`
This will build to `json-schema/bin/Release/net7.0/json-schema`
(json-schema was the old project name)

then from the root of ds repo run:  
`state-schema-gen/json-schema/bin/Release/net7.0/state-schema-gen core/dist/json-schema.json DawnSeekersUnity/Assets/Scripts/Cog/StateSchema.cs`

### DIFF
--- a/state-schema-gen/README.md
+++ b/state-schema-gen/README.md
@@ -7,7 +7,7 @@ It wasn't adding as dev dependency when I tried to do this and I'm not sure why
 Build core 
 I couldn't do this in isolation and had to build the whole game
 
-Run the json schema generator
+Run the json schema generator from the `core` folder
 `npx typescript-json-schema tsconfig.json DSState -o dist/json-schema.json`
 
 NOTE: If it failes due to bigint not being serialisable, You need to add the following decorator to the field
--- a/state-schema-gen/README.md
+++ b/state-schema-gen/README.md
@@ -4,23 +4,18 @@ Make sure `typescript-json-schema` is installed in `core`
 `npm install typescript-json-schema -g`
 It wasn't adding as dev dependency when I tried to do this and I'm not sure why
 
-Build core 
+Build core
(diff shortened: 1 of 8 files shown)


state-schema-gen/README.md
```diff
@@ -0,0 +1,21 @@
+# Codegenning the schema
+
+Make sure `typescript-json-schema` is installed in `core`
+`npm install typescript-json-schema -g`
+It wasn't adding as dev dependency when I tried to do this and I'm not sure why
+
+Build core
+I couldn't do this in isolation and had to build the whole game
```
chrisfarms, line 8: `make core/dist/core.js` should work (at least in new branch) .... but yeah just `make` is probably a safer bet

chrisfarms (approved):
👍

PR #44, opened 2023-03-24
Pending states for inventory transfers

Chris-Baker:
Add predictive prending states to inventory transfers
- Add the block number to GameState
- Store transfers in inventory provider
- Update transfers when a new block number arrives
- Add pending transfers to bags


chrisfarms (approved):
lovely stuff ... that's a much easier way of doing the optimisticness than I had in my head 👏 ... I shouldn't have any trouble fixing that up in my rebase

PR #46, opened 2023-03-27
Webgl improvements

JackPlaymint:
This PR adds a loading bar to the WebGL content in the shell to entertain our player's who enjoy rectangles that grow over time, while they wait for Unity to load up.

Once Unity has loaded, it's now masked with a solid color overlay that fades out once some map tiles have loaded in, just to add a more pleasant transition and to mask too much pop in.

I'm unsure if the css for the loading bar in unity-map/index.tsx is correct (like should it be in the styles script?) Couldn't work out how to do that, but let me know if it needs doin!


chrisfarms (approved):
> I'm unsure if the css for the loading bar in unity-map/index.tsx is correct (like should it be in the styles script?) 

I think the idiomatic thing to do would be to create a seperate `<Progress loading={x} />` component, and follow the same StyledComponent patten as the other components..... but since this is only being used in the one place, it's not the end of the world having it as you have done

PR #47, opened 2023-03-27
Map state restructure

hypnoshock:
# What

## Refactored map to use the new `GameState` object

- Removed the mapHack from the frontend
- Regenned the C# GameState schema
  - This wasn't a 100% success. Tiles and Seekers that are children of different object types are seen as totally different types in the C# end. You'll see `Seekers`, `Seekers3`, `Tiles` and `Tiles2`. I'm not sure how to elegantly fix this yet so there are places I have had to add overloads for each of the Types that are the same. This can be seen in `TileHelper` for instance.
- Renamed `PluginController` to `GameStateMediator` 
  - `PluginController` was a hang over from when the map was seen as a plugin but since we have shifted away from that I thought it made sense to rename the class to something more descriptive to what it does

## Other bug fixes

- Fixed problem where the player was able to right click on a fringe unscouted tile to complete the move path
- Fixed bug where action menu would stay present when a seeker had moved off the tile
  - This was because the state of the action menu was still local and not driven by `GameState`. That was a necessary hack when multiple tiles weren't part in the `GameState` back in the old days of 4 weeks ago

## Thing of note
- `typescript-json-schema` still uses TypeScript 4.8 and we use 4.9 therefore its version of `tsc` missing the `satisfies` keyword. Had to globally replace `satisfies` for `as` to make it work (reverted the change after I generated the json)
- Generated C# Schema doesn't allow `Buildings` or `Intent` to be null so I had to manually remove the nonNullable flag. Might be something we can do with the TypeScript interfaces to make that come through properly but I think somewhere the `partial` interfaces have been lost in translation

## Known problems
- `prevLocation` and `nextLocation` on the Seeker hasn't had any fields generated for C#!
  - I can see the fields in the json schema but for some reason the library that does the conversion of json-schemas to C# classes doesn't know how to use them. To make the map work I've had to manually dig out the fields using NewtonSoft's `JObject`. This can be seen in `TileHelper` where I find return seeker position.
  - Also the Seeker position code is in `TileHelper` as we used to be able to pass the tile found on `nextLocation` into the helper function.
- The seekers as seen on the map seem to be one step behind where they actually are. I have gotten around then problem for the player but using the coordinates as found in the player's seeker instead of checking if the seeker that belongs to a map tile is a player seeker or not. Although this has 'fixed' the problem, you can see still see that this is happening if you move on top of another seeker. It won't move out of the way to make room for the icon but if you then move your seeker to another tile the other player seeker will then move out of the way for a phantom seeker.
- Switching accounts doesn't reset the player's icon. Also you can see witness the lagging position problem where the player's seeker will move to it's previous position and the new account's seeker will move into position


chrisfarms:
Awesome 👏  .... spun it up, does what it says on the tin 👍 




> prevLocation and nextLocation on the Seeker hasn't had any fields generated for C#!

I think it's worth the time to get this code gen stuff working without manual intervention (doesn't have to be this PR)... I'm worried how fragile the interface between the map and everything else is, I can't imagine this is the last time we change bits of state queries and I'm concerned we don't get compile time errors.

Could we solve this by creating a dedicated `MapState` typescript type with exactly the shape that is best suited for the map?. This way we could ensure the type is friendly to the code generator, and we could use the type at the point of JSONification in the bridge/react code to ensure any changes to state stuff elsewhere is caught at compile time. This would decouple "what the map wants" from "whatever graphql we are building"

It's a bit of duplication of effort (since we wouldn't be taking advantage of the graphql generated types) but if it allows us to automate the generation and have type checking at the border I would consider it worth it.



> The seekers as seen on the map seem to be one step behind where they actually are...

OK I can see the problem now 🤦  

`world.tiles[].seekers[]` contains seekers at their _previous location_. See [./core/world.graphql:64](https://github.com/playmint/ds/blob/50f9a16bca78b97c14aeb38aa07ef614f30bc46a/core/src/world.graphql#L64) ... that `key:0` should be a `key:1`.

Screenshot 2023-03-28 at 07 43 30

Change that 0 -> 1, regen the graphql `npm run build:graphql -w core` and your one-step-behind will probably disapear.

Looking at the code around this ... it doesn't look like the map _really_ wants seekers attached to tiles _at all_. We could drop the `world.tiles[].seekers[]` .... and just have `world.seekers[]` 🤷  .... we can and possibly should change the shape of the world state into something the map finds useful, rather than taking my guess and making the map do extra work 




> Switching accounts doesn't reset the player's icon. 

I got nothing, on that one sorry

hypnoshock:
> Could we solve this by creating a dedicated MapState typescript type with exactly the shape that is best suited for the map?. This way we could ensure the type is friendly to the code generator, and we could use the type at the point of JSONification in the bridge/react code to ensure any changes to state stuff elsewhere is caught at compile time. This would decouple "what the map wants" from "whatever graphql we are building"

Agreed, I think this is the only way of doing it. 

> Looking at the code around this ... it doesn't look like the map really wants seekers attached to tiles at all. We could drop the world.tiles[].seekers[] .... and just have world.seekers[] 

That's more like it was before. Conceptually it also makes sense to decouple from the tiles. If we ever get to the point of lerping between the prev and next locations it really doesn't make sense to think of a seeker belonging to a tile

DawnSeekersUnity/Assets/Map/Scripts/MapInteractionManager.cs
```diff
@@ -97,9 +95,12 @@ void MapClicked()
         }
 
         // Select the tile
-        if (PluginController.Instance.WorldState.UI.Selection.Intent == Intent.NONE)
+        if (
+            GameStateMediator.Instance.gameState.Selected.Intent == null
+            || GameStateMediator.Instance.gameState.Selected.Intent == Intent.NONE
+        )
```
chrisfarms, line 101: Is this guard what is preventing selecting tiles when in the construct intent?

If the guard there so that it doesn't conflict with the path selection logic, then maybe it should be:

```suggestion
        if (GameStateMediator.Instance.gameState.Selected.Intent != Intent.MOVE)
```

Chris-Baker (approved):

PR #51, opened 2023-03-28
Drop single inventory items with right click

Chris-Baker:
- Drop single items with right click
- Show a pending animation on pending to slots
- Disable interaction on pending from slots
- Stop items in player hand from triggering scroll bars
- Clicking outside of a bag causes items in hand to reset

Still to come
Batching up transfers when the user clicks to drop a single item quickly

Note: I have captured context menu (right click) events at the top of the app. Use f12 to open dev tools instead of inspect


hypnoshock (approved):
That's a tick from me ✅

PR #52, opened 2023-03-28
core: wait for dry-run before dispatching next

chrisfarms:
before... pushes all actions to sequencer as fast as possible

after... waits for the action to get accepted (ie gets past dry-run) before dispatching next one in the queue

this gives a better experience when signing is required


hypnoshock (approved):
The power a 6 character change can bring

PR #54, opened 2023-03-28
Map scouting

hypnoshock:
# What

Selecting the scout intent button will highlight scoutable tiles in green
image

Multiple tiles can be selected which turns them orange
image

Like move, right clicking will execute the intent
image

When there are no more tiles to scout intent is set back to NONE (see the seeker's tile is no longer orange)
image

Oh and the fringe tiles are back again (sorry my b!)

## Details

- Added buttons for both Scouting and Construction intents
  - Used nouns project and added them to CONTRIBUTIONS.md 
- Renamed `SeekerMovementManager` to `MoveIntent` so it matches the naming of conventions of an intent
  - There is now an intent prefab in the scene that has an GO for each intent script 
- Moving between intents will clear the selection
  - I know it isn't quite what we want ultimately but because tiles selected for move and tiles selected for scout are mutually exclusive and we have just the one array to hold selected state things were starting to get messy so for now easier to clear them
- Scout intent will show all tiles that are valid for scouting in green
- These tiles can be selected which mutates `GameState` the same way as selecting a movement path does


DawnSeekersUnity/Assets/Map/Scripts/Intent/ScoutIntent.cs
```diff
@@ -0,0 +1,247 @@
...
+            // Clean up the selection - deselect tiles that have been scouted
+            var validSelectedTileIDs = selection
+                .Where(
+                    cellPosCube =>
+                        _validTilePositions.Contains(cellPosCube) || cellPosCube == _seekerPos
+                )
+                .Select(cellPosCube => TileHelper.GetTileID(cellPosCube))
+                .ToList();
+
+            GameStateMediator.Instance.SendSelectTileMsg(validSelectedTileIDs);
```
chrisfarms, line 77: sorry, you knew I was gonna pick on you for this 😆 

I would _really_ like it if we could avoid this. Consider if two different parts of the system were both triggering selections as soon as they recv state ... it's will end in an infinite loop of selection war. Ideally only explicit user interaction can change the selection.

Rather than altering the selection, instead can we just _never trust that the selection is valid_.

By that I mean, anything that wants to use the selection, should first filter out anything that doesn't make sense, and simply ignore everything else.

So for example, 

`GetSelectedTilePositions` in scout intend should not blindly trust the `state.Selected.Tiles`, and instead should return `state.Selected.Tiles.Where(isValidForScouting)` 

same in MoveIntent ... don't assume the Selection is valid, but pick what is valid from it.

I'm sure I'm missing some nuance in the implementation that probably makes that harder than I made it out.... but do think it's important to keep the selection handling clean as plugins, map, shell everything can mess with selections and things could feel real buggy if we are not strict.

Think of it this way: The selection is what the user thinks they want, not necasarily what they get. .... I might select a bunch of junk ids that are meaningless to the map, but maybe they have meaning to some other component.

chrisfarms (approved):

DawnSeekersUnity/Assets/Map/Scripts/Intent/ScoutIntent.cs
hypnoshock: Looks like you already aproved the PR but I've updated it to address this now.

PR #56, opened 2023-03-29
chore: getting frontend build ready for deploy

chrisfarms:
* adds some config to allow picking endpoint at build time (not ideal - runtime would be better - but a start)
* fixup some of the docker build context so it includes the wasm map and ignores ~2GB of context


hypnoshock (approved):
LGTM

PR #58, opened 2023-03-29
Construction intent implementation

hypnoshock:
# What

Implementation of construction intent along with some tweaks to not have the action menu rely on the seeker tile having to be the first in the selection

- Construction intent allows the player to select only 1 tile adjacent to the seeker if it doesn't have a building on it
- Removed quirk where the `ActionMenu` required the first tile in a selection to be the seeker tile
  - If we are in any of the handled intents then the action menu will be shown
- Removed the first tile being the seeker quirk from the `ScoutIntent`

## Known issues

There's a double of code across Move, Scout and Construct intent handlers. We could put some of that stuff either in the base `IntentHandler` class or make some kind of generic selection manager which handles this stuff so not each intent is handling what's highlighted. One for another day...

### DIFF
--- a/DawnSeekersUnity/Assets/Map/Prefabs/Intents.prefab
+++ b/DawnSeekersUnity/Assets/Map/Prefabs/Intents.prefab
@@ -9,7 +9,7 @@ GameObject:
   serializedVersion: 6
   m_Component:
   - component: {fileID: 6406227089287717616}
-  - component: {fileID: 7830148173117367100}
+  - component: {fileID: 5830282379756671830}
   m_Layer: 0
   m_Name: Intents
   m_TagString: Untagged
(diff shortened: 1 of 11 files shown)


chrisfarms:
> We could put some of that stuff either in the base IntentHandler class or make some kind of generic selection manager

I think the duplication is probably the right thing to do until we have some kind of final design .... for all we know the selections look and work totally differently to each other and a generic thing might end up a pain ... so this this all seems harmless to me

hypnoshock:
> I think the duplication is probably the right thing to do until we have some kind of final design

Cool, maybe I was over thinking it. A bonus of them all doing their own thing is it's easy to have the tiles be different colours for the different intents if we ever decide to do that

chrisfarms (approved):
very readable 10/10 would read again

PR #59, opened 2023-03-29
frontend: remove redundant construct button

chrisfarms:
we now have the nice map controls for entering the construct intent so can get rid of the temp button


hypnoshock (approved):
Button be gone

PR #62, opened 2023-03-29
Item transfer batching

Chris-Baker:
Add dispatch batching for transfers so we don't spam single item transfers.

Wait for 1 second before dispatching the transfers so that we can add extra items if we want to


frontend/src/plugins/building/index.tsx
```diff
@@ -167,7 +167,6 @@ const ConstructAvailable: FunctionComponent<ConstructAvailableProps> = ({ tile,
         e.preventDefault();
         const form = new FormData(e.target as any);
         const data = Object.fromEntries(form.entries());
-        console.log(data);
```
chrisfarms, line 170: 😇

chrisfarms (commented):
LGTM ... but if you are saying it's bit temperamental I'll leave it unticked until lewis has kicked it about a bit

ldunnplaymint (approved):

PR #87, opened 2023-04-03
Use slot key instead of array order

Chris-Baker:
Use slot key instead of array order to filter pending transfers

### DIFF
--- a/frontend/src/plugins/inventory/bag/index.tsx
+++ b/frontend/src/plugins/inventory/bag/index.tsx
@@ -7,6 +7,7 @@ import { BagFragment } from '@dawnseekers/core';
 import { FunctionComponent, useEffect, useRef } from 'react';
 import styled from 'styled-components';
 import { styles } from './bag.styles';
+import { getNewSlotKey } from '@app/plugins/inventory/helpers';
 
 export interface BagProps extends ComponentProps {
     bag: BagFragment;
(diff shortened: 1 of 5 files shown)


frontend/src/plugins/inventory/bag/index.tsx
```diff
@@ -28,42 +29,46 @@ export const Bag: FunctionComponent<BagProps> = (props: BagProps) => {
...
 
-    const slots: BagSlotProps[] = Array(numBagSlots)
-        .fill(null)
-        .map((_, slotIndex) => ({
+    // we need 4 slot keys so we are going to use any existing ones
+    // and then generate the remainder
+    const slotKeys = bag.slots.map((s) => s.key);
+    for (let n = 0; n < numBagSlots - bag.slots.length; n += 1) {
+        slotKeys.push(getNewSlotKey(bag, n));
+    }
```
chrisfarms, line 37: this feels like it might get confused in some pathalogical case where a key=10 has been created (which _shouldn't happen_ but there's nothing stopping it from happending yet).

would it not be easier to say that the inventory ONLY operates on slots 0,1,2,3 and just ignores anything else? ie:

```
const slotKeys = [0,1,2,3];
```

that way you wouldn't need to try and back fill any slots... your `slotKeys` would always be the fixed array which guarantees we are only ever dealing with those slots

Chris-Baker: This way allows us to use the first 4 keys regardless of their slot key value, and generates new keys starting at 0. 

So for example if we have keys `[203, 144, 10]` for some reason we still generate `0` as the next one and use that. In the usual case we still get slot keys `[0, 1, 2, 3]`

We could constrain the inventory as you say, but this should allow both

PR #101, opened 2023-04-05
chore: fix comment

### LINKED ISSUE #97: TRANSFER_ITEM_SEEKER Comments don't match variable names
TRANSFER_ITEM_SEEKER is an action used in a crafting extension so this is some user facing docs.
It's quite confusing as to what the various slot and bag indexes refer to.

https://github.com/playmint/ds/blob/593b009b899af0438fad6f4352a6c555fec359f6/contracts/src/actions/Actions.sol#LL19

chrisfarms:
closes #97

### DIFF
--- a/contracts/src/actions/Actions.sol
+++ b/contracts/src/actions/Actions.sol
@@ -16,7 +16,8 @@ interface Actions {
     // to scout location
     function SCOUT_SEEKER(uint32 sid, int16 q, int16 r, int16 s) external;
 
-    // transfers a balance between item slots
+    // transfer a qty of items from itemSlot[0] in equipees[0]'s equipSlots[0] bag
+    // to itemSlot[1] in equipees[1]'s equipSlots[1] bag
     // bags must be at same location as seeker
     // bags must either be owned by seeker or owned by nobody
     function TRANSFER_ITEM_SEEKER(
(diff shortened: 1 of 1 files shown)


contracts/src/actions/Actions.sol
```diff
@@ -16,7 +16,7 @@ interface Actions {
     // to scout location
     function SCOUT_SEEKER(uint32 sid, int16 q, int16 r, int16 s) external;
 
-    // transfers a balance of whatever is in fromBag/fromSlot into toBag/toSlot
+    // transfers a balance between item slots
```
5p0rt5BEArD, line 19: I would still have to go and read the code to see what params do. Its not clear what the size 2 arrays are.

how about:

// transfer a qty of items from itemSlot[0] in equipees[0]'s equipSlots[0] bag, to itemSlot[1] in  equipees[1]'s equipSlots[1] bag

5p0rt5BEArD (changes requested):
I appreciate this is a long winded way to change a comment - I was more dipping my toes in the issue system when I wrote the issue. But as a user it would have been useful to have a more explicit comment.

contracts/src/actions/Actions.sol
chrisfarms: done!

chrisfarms:
hah yeah def quicker to press the edit button that raise an issue 😆  ... but good practice

obviously the _real_ issue here is "have some proper documentation with examples" as I think similar comments could be made about pretty much every action being rather mysterious and opaque 😄

5p0rt5BEArD (approved):

Issue #88, opened 2023-04-03
Construction is unnecessarily fussy about inventory slots


PlaymintDavid:
I had enough wood to construct my first building, but the action failed until I put the wood in a certain place in my inventory.

<img width="304" alt="Screenshot 2023-04-03 at 14 11 08" src="https://user-images.githubusercontent.com/129759334/229520347-b5f6caca-cd10-4d77-b39a-dc3c8981161e.png">

chrisfarms:
Yep this is currently hacky. The "magically stealing stuff from a specific slot in your bag" thing is just a stop-gap until there is a UI to drag-n-drop the required resources. This issue also affects crafting, which also steals things from your bag.

We should not be having anything stolen from bags without explicit inventory dragy-droppy action.

PR #128, opened 2023-04-12
Construction drop target

Chris-Baker:
## What is this?

fixes: https://github.com/playmint/ds/issues/88

Add drop target to the building plugin ui
image

Add action to allow transferring items to the building bag before it is constructed

## What is missing?
Pending states for transfers into and out of the building bag
We have disabled the distance / location checks for resource transfers temporarily


PlaymintDavid:
Nice!

frontend/src/plugins/building/index.tsx
```diff
@@ -153,13 +162,13 @@ const ConstructAvailable: FunctionComponent<ConstructAvailableProps> = ({ tile,
         e.preventDefault();
         const form = new FormData(e.target as any);
         const data = Object.fromEntries(form.entries());
-        // TODO/FIXME: allow choosing where to pay from, not assume seeker/0/0
+
         player.dispatch({
```
chrisfarms, line 166: I don't think this dispatch matches the updated the CONSTRUCT_BUILDING_SEEKER action (too many args?)

Chris-Baker: Ah yes, updated. Thanks :)

PR #155, opened 2023-04-21
Add a dev build to maker file and clean up unity build script

ldunnplaymint:
### What?
I have added another maker command that calls a unity build script function to make a build but with development mode turned on

### Why?
Unity debug output by default is pretty rubbish so this should help finding and squashing bugs

### How?
I have copied the make map command and make it call a new unity build script function called DevBuild. In DevBuild, the unity options parameter is set to "Development" instead of "None". This does a few things but I have mainly done it for more verbose logging.

### So what?
If you do a make clean, a make debugmap and then a make dev, you will get a build with big verbose logging enabled

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,9 @@ all: node_modules contracts/lib/cog/services/bin/ds-node contracts/out/Actions.s
 map:
 	$(UNITY_EDITOR) -batchmode -quit -projectPath ./DawnSeekersUnity -executeMethod BuildScript.GitHubBuild -buildTarget WebGL -logFile - 
 
+debugmap:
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -26,6 +26,9 @@ all: node_modules contracts/lib/cog/services/bin/ds-node contracts/out/Actions.s
 map:
 	$(UNITY_EDITOR) -batchmode -quit -projectPath ./DawnSeekersUnity -executeMethod BuildScript.GitHubBuild -buildTarget WebGL -logFile - 
 
+debugmap:
```
chrisfarms, line 29: please add this new target to the list of .PHONY targets at the bottom of the file

(PHONY means "doesn't actually make a thing called debugmap")

chrisfarms (approved):

PR #164, opened 2023-04-26
3D map visuals

JackPlaymint:
This PR adds 3D seekers, bags and buildings into the map scene.
The map elements used to only consist of icons, and so the icons were used as the main controllable objects.
However that has now been swapped over to a new MapElementController class (and SeekerController which inherits), and the icons are now a more passive system that simply follows these map elements.

Additionally, the Seekers now have special behaviours when they share a tile with other seekers or when they are on a tile with a building. 

As well as the new systems, this PR also includes funky new art assets from Tom 'The Talent' Hoddle, including the new beaver statue and beaver head icons!


hypnoshock (approved):
Looks good to me. I haven't run it yet but the code looks alright. Good work! Looking forward to seeing those funky assets

PR #172, opened 2023-04-26
Only select tiles when not dragging

JackPlaymint:
This PR adds a check to only pass a select event if the user hasn't dragged the camera.
As a result it also changes the interaction from an OnMouseDown to an OnMouseUp event.


hypnoshock (approved):
Code looks alright but I haven't spun it up to actually test it

PR #194, opened 2023-05-03
Rotate the map

JackPlaymint:
In the beginning, the map was vertical, and it was good.
Then I forgot why it needed to be vertical, something to do with the tilemap system I think.
Now it no longer needs to be vertical as we've switched to 3D mode, and Lo, the map was made horizontal like a normal, sensible game.
This PR handles switching over to horizontal mode, swizzling the Vector3s in a bunch of map elements and effects.
It also updates some prefab orientations and positions.

Should make working on the scene more intuitive going forward.

### DIFF
--- a/DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
+++ b/DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
@@ -53,6 +53,7 @@ public class MapInteractionManager : MonoBehaviour
 
         if (Physics.Raycast(ray, out hit))
         {
+            Debug.Log(hit.transform.name);
             //Get the point that is clicked
             Vector3 hitPoint = hit.point;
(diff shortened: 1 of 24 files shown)


DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
```diff
@@ -53,6 +53,7 @@ private void Update()
 
         if (Physics.Raycast(ray, out hit))
         {
+            Debug.Log(hit.transform.name);
```
Chris-Baker, line 56: This might get noisy

Chris-Baker (approved):
lgtm

DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
JackPlaymint: Whoops! Yep, I'll get that now

PR #197, opened 2023-05-04
Increase pixel ratio for WebGL element

JackPlaymint:
Unity is a bit blurry in WebGL, this PR increases the resolution.
This could negatively affect performance on lower end platforms, so worth keeping in mind when we do our performance pass.

### DIFF
--- a/frontend/src/components/organisms/unity-map/index.tsx
+++ b/frontend/src/components/organisms/unity-map/index.tsx
@@ -189,7 +189,7 @@ export const UnityMap: FunctionComponent<UnityMapProps> = ({ ...otherProps }: Un
                     />
                 </div>
             )}
-            <Unity unityProvider={unityProvider} />
+            <Unity unityProvider={unityProvider} devicePixelRatio={2} />
         </StyledUnityMap>
     );
 };


ldunnplaymint (approved):

frontend/src/components/organisms/unity-map/index.tsx
```diff
@@ -74,6 +74,30 @@ export const UnityMap: FunctionComponent<UnityMapProps> = ({ ...otherProps }: Un
...
+            // device pixel ratio of the Unity Application when the device pixel
+            // ratio changes.
+            mediaMatcher.addEventListener('change', updateDevicePixelRatio);
+            return function () {
+                // Removing the event listener when the component unmounts.
+                mediaMatcher.removeEventListener('change', updateDevicePixelRatio);
+            };
+        },
+        [devicePixelRatio]
+    );
```
chrisfarms, line 99:
```suggestion
    useEffect(() => {
            // A function which will update the device pixel ratio of the Unity
            // Application to match the device pixel ratio of the browser.
            const updateDevicePixelRatio = function () {
                setDevicePixelRatio(window.devicePixelRatio);
            };
            // A media matcher which watches for changes in the device pixel ratio.
            const mediaMatcher = window.matchMedia(`screen and (resolution: ${devicePixelRatio}dppx)`);
            // Adding an event listener to the media matcher which will update the
            // device pixel ratio of the Unity Application when the device pixel
            // ratio changes.
            mediaMatcher.addEventListener('change', updateDevicePixelRatio);
            return function () {
                // Removing the event listener when the component unmounts.
                mediaMatcher.removeEventListener('change', updateDevicePixelRatio);
            };
        },
        [devicePixelRatio]
    );
```

chrisfarms (changes requested):

chrisfarms (approved):

PR #215, opened 2023-05-16
Editor intents window

JackPlaymint:
This PR adds a window to the editor that mimics the popup intents menu, as this will soon be replaced by ui in the shell.
At the moment I've left the popup intents menu in the scene as it is currently still needed in the WebGL build, but it is no longer required to play the game in the editor.

The 'Use' and 'Combat' intents have no effect at this time either, but they're ready for when we do.


chrisfarms (approved):
LGTM ... guess we'll need to do something smart about the "confirm" at some point to pass that along to the shell/bridge since I can't imagine we want to duplicate the confirm logic in two places (ie in EventTileRightClick() and in the shell-land) .... but looks easy enough to wire up when we get there

PR #229, opened 2023-05-24
New intents behavior

JackPlaymint:
This PR changes how engineer selection works.
From now on engineer selection is separate from tile selection.
Furthermore, you now no longer always have a engineer selected by default. You can deselect an engineer by clicking outside of its area of influence (one tile radius around the engineer).

The intents popup buttons that used to appear when you clicked on your seeker have been removed as they are now handled by Chris' shell UI intents buttons.
Consequently, the in-editor intents UI has been updated so that you cannot use the intents without a seeker selected.

This task had a lot more changes in the shell than I was expecting and I'm still not very comfortable reading and changing that code, so I'd really appreciate that bit being checked over in particular :)

### DIFF
--- a/DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
+++ b/DawnSeekersUnity/Assets/Map/Scripts/GameplayElements/MapInteractionManager.cs
@@ -136,6 +136,9 @@ public class MapInteractionManager : MonoBehaviour
                             )
                         )
                         .Contains(cellPosCube)
+                    && TileHelper.GetTilePosCube(
+                        GameStateMediator.Instance.gameState.Selected.Seeker.NextLocation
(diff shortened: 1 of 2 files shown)


ldunnplaymint:
Move intent button actice when no seeker selected, fires exception when clicked:`
client.js:1 KeyNotFoundException: The given key 'tile' was not present in the dictionary.`

chrisfarms (commented):
LGTM ... be good to remove the commented out stuff ... can always look into history if we need it another day

ldunnplaymint:
Blocking merge:

1. Yellow seeker
2. Move intent button active when random tile selected (Seeker not selected)
3. Selecting the seeker should always select the tile the seeker is on

ldunnplaymint (approved):

PR #239, opened 2023-05-26
Swap attack and defense param order in itemConfig to be correct

### LINKED ISSUE #235: Building Register helper functions get output item atoms wrong
If you declare an item with an output that contains attack atoms it checks the inputs defense atoms and vice versa.

This is down to the ItemUtils.register function getting the atom ordering wrong when creating the item kind node.

See https://github.com/playmint/ds/blob/fb481ef2aeb86b62b4bed7bc9d94cc2ecf78c1c8/contracts/src/utils/ItemUtils.sol#L41

5p0rt5BEArD:
Welcome drinks are now a little bit defensive rather than a little bit attacky. Should make for a safer, if less exciting, welcome party.

Will require HammerFactory to be fixed when this commit hits that repo's submodule.

resolves: #235

### DIFF
--- a/contracts/src/utils/ItemUtils.sol
+++ b/contracts/src/utils/ItemUtils.sol
@@ -11,8 +11,8 @@ struct ItemConfig {
     string name;
     string icon;
     uint256 life;
-    uint256 attack;
     uint256 defense;
+    uint256 attack;
     bool stackable;
     address implementation;
     string plugin;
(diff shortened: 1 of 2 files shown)


contracts/src/utils/ItemUtils.sol
```diff
@@ -38,7 +38,7 @@ library ItemUtils {
     // register is a helper to declare a new kind of item
     function register(BaseGame ds, ItemConfig memory cfg) internal returns (bytes24) {
         Dispatcher dispatcher = ds.getDispatcher();
-        uint32[3] memory outputItemAtoms = [uint32(cfg.life), uint32(cfg.attack), uint32(cfg.defense)];
+        uint32[3] memory outputItemAtoms = [uint32(cfg.life), uint32(cfg.defense), uint32(cfg.attack)];
```
chrisfarms, line 41: there's some constants for the indices ... maybe we should use them to populate this array to reduce chance of getting tripped up if/when we change atoms:

```
        atoms[ATOM_LIFE] = uint32(cfg.life);
        atoms[ATOM_DEFENSE] = uint32(cfg.defense);
        atoms[ATOM_ATTACK] = uint32(cfg.attack);
```

PR #267, opened 2023-06-06
fetch "model" building kind annotation

chrisfarms:
adds "model" annotation to the building kind query - we will use this to pick which 3d model (building vs enemy) to show in the map


hypnoshock (approved):
Approved

PR #273, opened 2023-06-07
Add combat UI

Chris-Baker:
Add combat UI
Clicking on "Start Combat" opens up the combat modal
image

Pre combat summary is currently missing, clicking "Start Combat" starts the combat session
image

Combat ticks down until one side wins
image

Clicking on "End Combat" finalises the session allowing players to collect their rewards
image

Combat rewards for the latest session are shown at the bottom of the screen
image


ldunnplaymint (changes requested):
The combat modal and the preview flicker every time there is a new tick

The combat rewards UI element is always visible with a combat selected, probs should't appear until there are rewards

PR #305, opened 2023-06-12
Pauls world stuff

PlaymintPaul:
Added lots of stuff to the map.


PlaymintPaul:
I see it's failed a the check-contracts test. Not sure why. Works fine on my local

chrisfarms (approved):
think I'd like to see some basic tests for this stuff at some point, since it's quite a lot of code to maintain ... not blocking this PR on that tho as probably not your responsibility :)

PR #306, opened 2023-06-13
Instantiating enemy models on map

hypnoshock:
# What

Any buildings with the model annotation of `enemy` will instantiate the enemy model instead of the typical building model

image


chrisfarms (approved):
😈

PR #311, opened 2023-06-13
Destroying buildings/enemies drop the recipe materials

hypnoshock:
Instead of dropping the equivalent of Kikis etc to match the atomic count of a building/enemy we are now dropping the materials that were used to build the building. This has the effect of enemies now dropping gold coins as designed.

Bling Bling
image

### DIFF
--- a/contracts/src/rules/CombatRule.sol
+++ b/contracts/src/rules/CombatRule.sol
@@ -209,7 +209,7 @@ contract CombatRule is Rule {
 
             CombatState memory combatState = calcCombatState(sessionUpdates, sortedListIndexes, ctx.clock);
             if (combatState.winState != CombatWinState.NONE && !state.getIsFinalised(sessionID)) {
-                _finaliseSession(state, combatState, sessionID, ctx);
+                _finaliseSession(state, combatState, sessionID);
             }
         }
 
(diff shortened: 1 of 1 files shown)


contracts/src/rules/CombatRule.sol
```diff
@@ -267,42 +259,34 @@ contract CombatRule is Rule {
...
+
+                for (uint8 j; j < 4; j++) {
+                    (bytes24 item, uint64 qty) = state.getMaterial(state.getBuildingKind(buildingState.entityID), j);
+                    // NOTE: divide by 200 instead of 100 if we want to halve the rewards as per the combat spec
+                    state.setItemSlot(
+                        Node.RewardBag(sessionID, winnerStates[i].entityID),
+                        j,
+                        item,
+                        uint64((damagePercent * qty) / 100)
+                    );
```
chrisfarms, line 281: should this check if the material item is stackable (or at least if there is qty>1) before attempting to split it up?

hypnoshock: I guess it would save gas? I just thought it would result to zero. Even if we did that check, if the damagePercent was too low to yield 1 unit then it would be zero anyway? I'm not dividing by zero so isn't it o.k?

PR #315, opened 2023-06-14
Add pre combat summary

Chris-Baker:
## What is here
image

## Known issues
- There are state transition issues after starting combat, the summary is removed for the first block of the combat
- The summary still uses the entity ID for building names during combat


hypnoshock (approved):
Surprised that seekers weren't already part of the selected tile data. LGTM
I like the bonux fix that stackable items no longer show quantity! Couldn't differentiate them before, nice one!

PR #332, opened 2023-06-20
Add under construction model

JackPlaymint:
This PR adds in the under construction model when you have started creating a building, but haven't confirmed the build yet.

the building should appear after you place the first element into the construction inventory.
It should remain there until either all placed elements have been removed from the construction inventory, or the confirm construction button has been pressed.

Also in this update is a small performance boosting measure that should help with the way that Map elements are handled.


DawnSeekersUnity/Assets/Map/Scripts/Environment/MapManager.cs
```diff
@@ -86,6 +86,14 @@ private void OnStateUpdated(Cog.GameState state)
     {
         foreach (var tile in state.World.Tiles)
         {
+            // If you can think of a better way of getting incomplete buildings on a tile, I'd like to hear it...
+            Cog.Buildings incompleteBuilding = state.World.Buildings.FirstOrDefault(
+                b =>
+                    b.Id.Contains(tile.Id.Substring(10))
+                    && b.Bags.Any(n => n.Bag.Slots.Any(s => s.Balance > 0))
+            );
```
chrisfarms, line 94: non-blocking: mentally noting this tight loop within a loop that is blocking the main thread for next week's performance work 😅  .... we probably need to precompute a map of incomplete-buildings-by-tile-id or something to avoid this

PR #333, opened 2023-06-20
Fix transitions between combat modal states and health bar rendering

Chris-Baker:
- Stop block time updates causing a complete DOM redraw.
- Smooth out the transition between pre-combat and combat


ldunnplaymint (changes requested):
On a fresh boot, the combat modal will not update after initiating a combat

1. Fresh boot/refresh
2. Make sure you have a valid session key
3. Move to an enemy/building
4. Attack and keep the modal up (Do not dismiss)
5. Note the modal does not update.


https://github.com/playmint/ds/assets/99176798/859aced1-421b-49e1-a6ef-03fd6fcb42b3

PR #334, opened 2023-06-20
Map performance pass

JackPlaymint:
This PR changes up the way that Tiles and Map elements are tracked and accessed, reducing garbage and increasing speedy fast quickness and also zoom zooms.

It touches a lot of functionality, anything that wants to know about a tile at a location.
Hopefully it should make getting information from the shell a bit quicker.

More work to do, but this is a start.

...If it doesn't work blame chatGPT as always...


ldunnplaymint (approved):
Did all the actions at least once, no errors or anything. Scouted a fair bit, held up. Needs hammering with more people but approved for main.

Issue #79, opened 2023-04-03
no way to get status of a dispatched action


chrisfarms:
dispatches go into a queue with several stages ranging from "queued locally in the client" through to "confirmed/failed on chain" .... but there is currently no way exposed for components to get the current status of where a dispatched action is within the queue.

we need to:

* monitor the status of actions being processed by the router/sequencer to gather the statuses
* have dispatch() return a unique `id` for the dispatch call
* expose a way to subscribe to the state changes for a given `id`

Issue #80, opened 2023-04-03
design and implement movement rule onchain


chrisfarms:
movement rules are current entirely enforced by the client ... which is no enforcement at all

movement needs to account for:

* complex movement paths
* blockers / tiles existing
* travel time
* ...?

PR #336, opened 2023-06-21
optimistic action simulation

chrisfarms:
## ⚠️ BEFORE MERGE

* [x] merge https://github.com/playmint/cog/pull/20
* [x] pin to the correct lib/cog 
* [x] rebase/squash

## What

updates to make use of the optimistic state updates so that we can have a significantly slower block time while still seeing fast state updates from everyone connected to the same indexer.

* we now subscribe to "blocks" not "state events" which reduces the amount of superfluous state fetches
* we fetch state from the simulated state (which updates as a fast as it can)
* simulated state gets merged with real state every 10s - so any reorgs of state should occur after 10s ish
* `dispatch()` now returns a promise that when resolved means state has successfully applied (resolves: #79 )
* updates all gql to fetch simulated state
* error revert reasons will now be displayed in the log
* various query tweaks to tune things against the slower dispatch / faster state updates
* contracts: replace all custom errors with error strings so that we can decode them without abi
* contracts: movement travel time is disabled (matching what the client expects) as until the client can handle travel time (see #80), the longer block time will break movement
* deployment: deployment updated to provision dedicated disks for cog services to all anvil some scratch space
* deployment: deploy an anvil process in the cog services pod to use as the simulation chain

## Bonus

* adds an additional "ds-exp" deployment to use for PR testing without clobbering "ds-test"
* builds a cog services image specifically for each deployment so that we can test with different pinned cog versions

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -22,6 +22,11 @@ on:
         required: true
         default: playmint/ds-contracts
         type: string
+      SERVICES_IMAGE_NAME:
+        description: ghcr image package name for cog services
+        required: true
+        default: playmint/ds-services
+        type: string
       PLATFORMS:
         description: platforms for multiarch images
         required: true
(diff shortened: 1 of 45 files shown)


contracts/test/rules/MovementRule.t.sol
```diff
@@ -79,58 +78,58 @@ contract MovementRuleTest is Test {
...
-        // we should now have arrived at -1,2,1
-        assertEq(
-            state.getCurrentLocation(seeker, uint64(block.number)),
-            Node.Tile(DEFAULT_ZONE, -1, 1, 0),
-            "expected arrive at -1,1,0"
-        );
-        // stop being alice
-        vm.stopPrank();
-    }
+    // function testMoveWhenMoving() public {
```
chrisfarms, line 81: this test is disabled until we are ready to enable movement travel time

.github/workflows/build.yaml
```diff
@@ -200,12 +205,68 @@ jobs:
...
+            VERSION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.version'] }}
+            REVISION=${{ fromJSON(steps.meta.outputs.json).labels['org.opencontainers.image.revision'] }}
+
   deploy:
     name: Deploy
     runs-on: ubuntu-latest
     needs:
     - frontend
     - contracts
+    - services
```
5p0rt5BEArD, line 269: How were things working if services not part of the ds build before?

core/src/cog.graphql
```diff
@@ -18,21 +18,11 @@ mutation dispatch($gameID: ID!, $actions: [String!]!, $auth: String!) {
...
-            skey: key
-        }
-        ... on RemoveEdgeEvent {
-            rfrom: from
-            rrel: rel
-            rkey: key
-        }
-        ... on SetAnnotationEvent {
-            afrom: from
+        ... on BlockEvent {
```
5p0rt5BEArD, line 23: In affect, subscribing to every block in simulation mode is subscribing the result of avery transaction made because its automine?

And this is les frequent than before because there may be multiple events per transaction?

frontend/src/components/organisms/unity-map/index.tsx
```diff
@@ -120,9 +120,13 @@ export const UnityMap: FunctionComponent<UnityMapProps> = ({ ...otherProps }: Un
 
     const newMapBlob = JSON.stringify(game);
     if (isReady && game && globalLastBlob != newMapBlob) {
+        console.time('mapsend');
```
5p0rt5BEArD, line 123: is this some profiling left in by mistake or here to stay on purpose?

5p0rt5BEArD (approved):
minor Qs

PR #340, opened 2023-06-21
fix for missing buildings

JackPlaymint:
The buildings are missing, technically that's good for performance, but ultimately it negatively effects gameplay slightly.
After much careful deliberation and pressure from my peers, I have put them back in.


chrisfarms (approved):
> technically that's good for performance

😂

Issue #301, opened 2023-06-12
Selecting an intent should deselect any selected tiles


ldunnplaymint:
Really annoying when I click move and it keeps the last tile I had selected as the first step in my move action.

PR #342, opened 2023-06-21
deselect on intent switch

chrisfarms:
dont trigger selectTiles on intent switch when switching from nothing->something

otherwise deselect tiles when switching intents

this reduces the anger in lewis

fixes: #301


ldunnplaymint:
This doesn't stop the one case I hate


https://github.com/playmint/ds/assets/99176798/0f3cf0e4-5f5a-4745-86e8-882d3d31ac58

ldunnplaymint:
Fair enough but the body of that issue is explicit about the Tile selected > Move intent flow. I think letting go of selected tiles between intents should still go in but I don't think it marks that issue as done.

chrisfarms:
I think you need to wait for "selecting bags and buildings" to resolve the issue you hate ... because presumably the reason you are selecting that tile in your vid is that you wanted to look into the bag or something

ldunnplaymint (approved):

PR #387, opened 2023-06-26
send state JSON to map into smaller chunks

chrisfarms:
### what

Instead of sending a single giant blob of JSON on each state update to the map, send many smaller chunks and reassemble them.

### why

deserialising json in unity-land is synchronous (and their doesn't appear to be a useful async api), and js is a single thread, which means deserialising the large blob of json blocks the event loop.

splitting up the sending of data to the map into smaller bits means that it takes significantly longer overall, but will yield cpu to the event loop in-between chunks, making things a bit more responsive.

### notes

ngl, I don't love this.... it def avoids the large blocking calls, but at the cost of a fair bit more complicated, more fragile code .... changes to the graphql may require equivilent changes to the sending code, it's quite a bit slower and increases the chances of things getting out of sync.

### DIFF
--- a/DawnSeekersUnity/Assets/Scripts/Cog/Plugin/GameStateMediator.cs
+++ b/DawnSeekersUnity/Assets/Scripts/Cog/Plugin/GameStateMediator.cs
@@ -116,7 +116,8 @@ namespace Cog
                     }
                     catch (Exception e)
                     {
-                        Debug.Log($"GameStateMediator::Update() error {e.Message}");
+                        Debug.Log("GameStateMediator::Update() error");
+                        Debug.LogError(e);
                     }
                 }
             }
@@ -178,7 +179,7 @@ namespace Cog
 
         public void NodeProcessThread()
         {
-            if (DawnseekersDevSettings.instance.NodePath == "") 
+            if (DawnseekersDevSettings.instance.NodePath == "")
             {
                 Debug.LogError("PluginController: Node path not set. Make sure the absolute path to node is set in the Edit > Project Settings > Dawnseekers panel");
                 return;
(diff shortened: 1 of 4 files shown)


ldunnplaymint (changes requested):
I hit this when trying to build
```
> tsc --build --verbose

[09:03:24] Projects in this build: 
    * tsconfig.json

[09:03:24] Project 'tsconfig.json' is out of date because output file 'dist/index.js' does not exist

[09:03:24] Building project '/Users/lewisdunn/Repos/ds/bridge/tsconfig.json'...

src/index.ts:61:18 - error TS2554: Expected 6 arguments, but got 3.

61     const game = makeGameState(player, world, selection);
                    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

  ../core/dist/core.d.ts:2596:129
    2596 declare function makeGameState(player: Source<ConnectedPlayer | undefined>, world: Source<World>, selection: Source<Selection>, selectTiles: Selector<string[] | undefined>, selectSeeker: Selector<string | undefined>, selectIntent: Selector<string | undefined>): Source<GameState>;
                                                                                                                                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    An argument for 'selectTiles' was not provided.


Found 1 error.

npm ERR! Lifecycle script `build` failed with error: 
npm ERR! Error: command failed 
npm ERR!   in workspace: dawnseekersbridge@1.0.0 
npm ERR!   at location: /Users/lewisdunn/Repos/ds/bridge 
make: *** [bridge/dist/index.js] Error 1
```

chrisfarms:
> I hit this when trying to build

alway forget the bridge ... fixed

ldunnplaymint (approved):

PR #394, opened 2023-06-27
Seeker button is disabled during the spawn dispatch

hypnoshock:
## What

image

Spawn button disabled whilst dispatch taking place

## Why

To prevent the player from spawning multiple seekers

## How

- Updated the dispatch function return type to be `promise<void>` to reflect the changes to the update function. With the correct type it can be awaited
- Disabling the button during the dispatch and reenabling after the promise either fulfils or errors

## Problems

There is a moment where the promise is fulfilled and the button renabled before the newly spawned seeker has been fetched which leaves a window of opportunity to press it again. The dispatch errors are being caught by core so I can't renable solely on error and as the dispatch returns void, I don't know if the dispatch was a success or not. A work around could be just to not renable the button at all after the dispatch but if there was an error then the player would have to reload the game.

### DIFF
--- a/core/src/plugins.ts
+++ b/core/src/plugins.ts
@@ -63,7 +63,7 @@ export function makeAvailablePlugins(client: Source<CogServices>) {
  * when a plugin attempts to call dispatch but there is no connected player to
  * dispatch for.
  */
-function noopDispatcher(...actions: CogAction[]) {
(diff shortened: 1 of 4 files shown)


chrisfarms:
could you add the `resolves: xxx` bit to the PR description to link the original issue 🙏

ldunnplaymint (approved):

PR #405, opened 2023-06-27
Bags show up on defeated tiles if the player has a reward

hypnoshock:
## What
image

Showing a bag on a combat tile if there are rewards belonging to the player

## How

We check if the tile has a session and if the player has any rewards associated with that session. 

## Issues

Known issue is that we don't know without calculating the full combat state which side won so for now I'm going to assume that if there are rewards that it would have been the defenders that had lost. Bags belong to the session not the tile so in reality the bag could be shown on either tile


chrisfarms (approved):

chrisfarms:
> Exception fires when doing the following...

this is present in `main` I don't think it should block this PR

ldunnplaymint:
> > Exception fires when doing the following...
> 
> this is present in `main` I don't think it should block this PR

Is it? I couldn't reproduce it? If that's true, merge it

chrisfarms:
Screenshot 2023-06-27 at 17 00 20

yep just reproduced ... it's more of a general issue with account switching and the modal than combat

PR #406, opened 2023-06-27
MapElements now stick to tile (including spawned bags)

JackPlaymint:
This PR makes all map elements stick to the tiles that they're on so that they move with the tile animation.
So all buildings, enemies, etc will bounce in when the map first loads up.
Discovered bags will bounce in with their revealed tiles.


chrisfarms (approved):

ldunnplaymint:
Is this correct?

https://github.com/playmint/ds/assets/99176798/27cd6343-d52e-45f9-b238-a6feae3535f6

PR #422, opened 2023-06-28
fudge: give the plugin sandbox more memory

chrisfarms:
we hit the memory limit of the sandbox due to the large size of the state. increasing this number is not a realistic solution long term, but it'll buy us a bit more time while we sort out the state size problem


Chris-Baker (approved):
image

PR #423, opened 2023-06-29
Add new Favicon icon

ldunnplaymint:
I have added a new favicon icon from Jonny. 

Not sure if I did it right but I renamed his PNG to an ICO and chucked it in the repo with the same name as the old one.

I can see it: 
Screenshot 2023-06-29 at 12 19 47

Probs @Chris-Baker to review


chrisfarms (changes requested):
it should not be named ico ... ico is a different file format and will act funny if returned with wrong mime type .... we probably need to generate the multiple formats and make sure it's the right multiple of 48px

Chris-Baker:
> 

I made some adjustments to this @chrisfarms

chrisfarms (approved):

PR #427, opened 2023-06-30
Added chance of spawning multiple resources on one tile

hypnoshock:
## What

I have added two extra rolls to spawn resources on a tile. There is a 50% chance of spawning the next resource and then a 25% chance of spawning yet another. 

image

Seperate rolls as in the 2nd roll happens even if the 1st was unsuccessful.

## Why

To speed up progression in the game by having more resources on the map


chrisfarms (approved):
LGTM

PR #460, opened 2023-07-07
Example docs correction

5p0rt5BEArD:
Example pizza resoures 6 flasks of red, not 4


PlaymintPaul (approved):
Yeah. Not sure what happened there. Good find!

Issue #384, opened 2023-06-26
The user should be able to merge stacks on drop


PlaymintDavid:
The user should be able to complete stacks by dropping the same resource on top of them in their own bags: If there is a stack of 90 and I drop 50 on to it, the 90 becomes 100 and I have 40 left on my cursor

ldunnplaymint:
Currently you cannot stack stackable items into stacks of over 100, reworded to make clear we need to have better error states

PlaymintDavid:
Thank you.

PR #469, opened 2023-07-11
drop up to the stack limit onto partial stacks

chrisfarms:
* dropping a stack of 100 onto a slot containing 50 of the same item will now drop 50 into the slot and leave you with 50 in your hand.
* highlights for where you can drop reflect being able to drop onto partial stacks.
* skips attempting transfer to same slot or when qty==0
* resolves: #384


ldunnplaymint (approved):

hypnoshock (approved):
LGTM. That routine where you chunk the ID into an array of 4 byte numbers to get at the item properties would be handy in an item helper some day.

chrisfarms:
> LGTM. That routine where you chunk the ID into an array of 4 byte numbers to get at the item properties would be handy in an item helper some day.

yeah I think there's 3 copies of it in the codebase already

PR #477, opened 2023-07-13
New BagRule introduced

hypnoshock:
## What

A new game rule named `BagRule` which implements bag actions.

```
    function TRANSFER_BAG_OWNERSHIP(bytes24 bag, bytes24 fromEntity, bytes24 toEntity) external;

    function TRANSFER_BAG(bytes24 bag, bytes24 fromEntity, bytes24 toEntity) external;

    function SPAWN_EMPTY_BAG(bytes24 equipee) external;
```
## Why

Transferring of bags and ownership was done by directly manipulating the game state. For building implementation code this will be illegal therefore basic rules for bags needed to be made to expose this functionality

## Summary of the rules

- An owner of a bag can transfer a bag onto another `MobileUnit` or a `BuildingInstance`
  - Transferring a bag onto another entity does NOT inherently transfer ownership. This means it is possible to transfer a bag onto another entity and its use would be restricted from the point of view of that entity i.e. they cannot move items out of that bag nor transfer ownship
- Equipees of a bag that doesn't belong to them are allowed to transfer the bag onto another `MobileUnit` or a `BuildingInstance`
- An owner of a bag can transfer ownership of a bag to another `MobileUnit` or a `BuildingInstance` 
- Bags cannot be transferred directly to or from tiles.

## Problems encountered

The code that checks if a bag belongs to the `sender` of a action is a little tricky because of the way that bags are owned by `Player` nodes when equipped to `MobileUnits` but are owned by `Building` nodes (Instances of buildings) when equipped to a building. 

Trouble with validating that the sender was the owner of the bag when called from building contract

### DIFF
--- a/contracts/src/rules/BagRule.sol
+++ b/contracts/src/rules/BagRule.sol
@@ -1,0 +1,192 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import {State} from "cog/State.sol";
+import {Context, Rule} from "cog/Dispatcher.sol";
+
+import {Schema, Node, BiomeKind, Kind} from "@ds/schema/Schema.sol";
+import {TileUtils} from "@ds/utils/TileUtils.sol";
+import {Actions} from "@ds/actions/Actions.sol";
+import {BagUtils} from "@ds/utils/BagUtils.sol";
+
+using Schema for State;
+
+uint8 constant MAX_EQUIP_SLOT_INDEX = 254; // There appears to be a problem with state.getEquipSlot if used with 255
+
+contract BagRule is Rule {
+    function reduce(State state, bytes calldata action, Context calldata ctx) public returns (State) {
+        if (bytes4(action) == Actions.SPAWN_EMPTY_BAG.selector) {
+            // decode the action
+            (bytes24 equipee) = abi.decode(action[4:], (bytes24));
+
+            _spawnEmptyBag(state, ctx.sender, ctx.clock, equipee);
+        }
+
+        if (bytes4(action) == Actions.TRANSFER_BAG_OWNERSHIP.selector) {
+            // decode the action
+            (bytes24 bag, bytes24 toEntity) = abi.decode(action[4:], (bytes24, bytes24));
+
+            _transferBagOwnership(state, ctx.sender, bag, toEntity);
+        }
+
+        if (bytes4(action) == Actions.TRANSFER_BAG.selector) {
+            // decode the action
+            (bytes24 bag, bytes24 fromEntity, bytes24 toEntity) = abi.decode(action[4:], (bytes24, bytes24, bytes24));
+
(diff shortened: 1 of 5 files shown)


contracts/src/rules/BagRule.sol
```diff
@@ -0,0 +1,192 @@
...
+        state.setEquipSlot(equipee, equipSlot, bag);
+    }
+
+    function _transferBagOwnership(State state, address sender, bytes24 bag, bytes24 toEntity) private {
+        _requireSenderIsOwner(state, sender, bag);
+
+        _setOwner(state, bag, toEntity);
+    }
+
+    function _setOwner(State state, bytes24 bag, bytes24 owner) private {
```
chrisfarms, line 73:
```suggestion
    function _setOwner(State state, bytes24 bag, bytes24 entity) private {
```

contracts/src/rules/BagRule.sol
```diff
@@ -0,0 +1,192 @@
...
+        // Building kinds are directly set as owners whereas with MobileUnits we use the Unit's owner which is a Player
+        if (bytes4(owner) == Kind.Building.selector) {
+            state.setOwner(bag, owner);
+        } else {
+            state.setOwner(bag, state.getOwner(owner));
```
chrisfarms, line 78: handle the case for public (no owner / 0x0)

contracts/src/rules/BagRule.sol
```diff
@@ -0,0 +1,192 @@
...
+    function _transferBag(State state, Context calldata ctx, bytes24 bag, bytes24 fromEntity, bytes24 toEntity)
+        private
+    {
+        (uint8 equipSlot, bool found) = _getEquipSlotForEquipment(state, fromEntity, bag);
+        require(found, "BagRule::_transferBag: Bag is neither equipped to entity or owned by the entity");
+
+        _transferBagAtSlot(state, ctx, equipSlot, fromEntity, toEntity);
+    }
+
+    function _transferBagAtSlot(
```
chrisfarms, line 91: might be good if we can be explicit about the "to equip slot" rather than attempting to detect a free/empty one

contracts/test/rules/BagRule.t.sol
```diff
@@ -0,0 +1,346 @@
...
+
+import {Game as Downstream} from "@ds/Game.sol";
+import {Actions} from "@ds/actions/Actions.sol";
+import {Schema, Node, Rel, LocationKey, Kind, BiomeKind, DEFAULT_ZONE} from "@ds/schema/Schema.sol";
+import {ItemUtils} from "@ds/utils/ItemUtils.sol";
+import {BuildingKind} from "@ds/ext/BuildingKind.sol";
+
+using Schema for State;
+
+contract BuildingRuleTest is Test {
```
chrisfarms, line 18:
```suggestion
contract BagRuleTest is Test {
```

chrisfarms (approved):

ldunnplaymint (approved):

PR #482, opened 2023-07-27
Map element selection

JackPlaymint:
This PR adds selectable bags and buildings, instead of having to select the tile (although it does the same thing and selects the tile anyway). 
The bags and buildings have a nice outline effect that matches the Mobile Unit's.

The USE intent has been removed entirely, selecting a building while having a mobile unit selected will automatically 'use' that building.

You can now use a building you are standing on again.


ldunnplaymint:
I think selecting a tile object when a building/bag is present should still highlight the building/bag as they are both active in the side bar.

JackPlaymint:
> 

I think I agree, but when I spoke to Tom about this he was unsure. I think he wanted to try it out and se what it felt like and change it if it seemed weird. Happy to change it now or we could hang on for Tom on Monday?

ldunnplaymint:
Let's wait for Tom and Paul but make sure we point that quirk out.

chrisfarms (approved):
codewise all LGTM 👍 ... and it's all working with respect to the removal of the "use" intent

but I'm not totally sure I get the desired outcome with the element selection, It kinda feels like we are in between two different UX directions:

1. A world where you can select a single {Tile OR Bag OR Building} and only that thing highlights, and only data about that thing is shown
2. A world where you can select a Tile, and everything on it hightlights, and data for all things on that tile are shown

right now it acts like 2 ... but kinda allows a version of 1 where you can select a tile + a thing, but selecting the thing directly doesn't actually have any effect other than toggling the highlight, and where the tile is always selected.

If direction (2) is where we are heading ... then is `selectMapElement` selector stuff actually required? since it is kinda duplicating selectTile

JackPlaymint:
> 

Yeah I see what you mean. When we discussed this during sprint planning, I think what we landed on was that we would probably want map elements to be individually selectable, separate from the tile. Which would mean we would need to remove the Building and Bag bits out of the Tile class and move them into their own thing.

However, as Johnny will be redesigning the UX flow soon, we decided not to change up the current functionality too much as at this stage as it might all be overwritten eventually.

chrisfarms:
gottit 👍  ... so:

* The trajectory is towards (1) - where we can select individual things
* This PR basically contains steps towards that
* ...but we are not quite there yet and happy to sit in the middle ground between (1) and (2) until we have a confirmed design

In that case I think I just echo @ldunnplaymint 's point that we should probably just highlight _everything_ on the tile for now, to avoid the situation where it looks like you are selecting/deselecting something but it is having no effect other than showing/hiding the highlight .... guess wait for feedback from paul/tom 

it's fine to get this merged in the meantime imo, it's quirky but sounds like it's working as intended ... and it'll be easier for everyone to have a play from the main build / unblock @hypnoshock rebasing

ldunnplaymint:
I reckon make it select everything and then merge it. I think we will get complaints otherwise.

ldunnplaymint (approved):

Issue #485, opened 2023-07-31
automate ds-cli npm publish


chrisfarms:
publishing the ds-cli to npm is currently a manual process and requires an OTP code and a version bump... this will get stale often ... we should make a deployment key and github workflow to handle version bump + publish

ldunnplaymint:
Or compile it and publish it as a zip file

PR #486, opened 2023-07-31
ds-cli: basic cli skeleton

chrisfarms:
initial project layout and command structure for a downsteam cli 

this doesn't do anything very interesting yet, but gets all the bits in roughly the right shape ready to work on the building deployment bits.

I chose the [yargs](http://yargs.js.org/) lib to help with the structuring mostly because it was very easy to get going with, but it isn't great at typesafty (or at least I haven't managed to convince it to be yet) - so it's a little fast and loose with types.

the general command structure is:

```
ds [noun] [verb] [params]
```

I've made a couple of example commands to test we can read from the graphql...

```
ds items list
ds items search goo
ds items show <id>
```

... and dispatch actions

```
ds actions list
ds actions dispatch SPAWN_MOBILE_UNIT <id>
```

You can point the cli at different deployments by using the `--network` (or `-n`) option:

```
ds --network devnet items list
```

To see the available network endpoints use `ds networks`

```
ds networks list
```

Actions that require a signature/session will prompt to signin with a walletconnect QR code rendered in the terminal.

Outputs are roughly rendered using `console.table` for now .... or you can switch to JSON with the `-o json` option:

```
ds -o json items list
```

... I have tested publishing the package, but we will need to automate it ... created #485 to track this ... for now it can be published by hand via `npm run build && npm publish`

### DIFF
--- a/cli/.eslintrc.json
+++ b/cli/.eslintrc.json
@@ -1,0 +1,24 @@
+{
+  "plugins": ["@typescript-eslint"],
+  "extends": [
+    "next/core-web-vitals",
+    "plugin:@typescript-eslint/recommended",
+    "plugin:prettier/recommended",
+    "prettier"
+  ],
+  "rules": {
+    "@typescript-eslint/explicit-module-boundary-types": "off",
+    "@typescript-eslint/explicit-function-return-type": "off",
+    "@typescript-eslint/no-explicit-any": "off",
+    "@typescript-eslint/no-object-literal-type-assertion": "off",
+    "@typescript-eslint/no-var-requires": "off",
+    "@typescript-eslint/camelcase": "off",
+    "@typescript-eslint/ban-ts-comment": "off",
+    "@typescript-eslint/no-empty-function": "off",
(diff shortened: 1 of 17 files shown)


ldunnplaymint:
I hit this when trying to do the tile deploy on windows:
```SyntaxError: Unexpected token ' in JSON at position 0
    at JSON.parse (<anonymous>)
    at Object.handler (C:\Users\lewis\AppData\Roaming\npm\node_modules@playmint\ds-cli\dist\ds.js:88158:18)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)```

chrisfarms:
is there any more output? looks like that has been truncated so cant see the error

chrisfarms:
@ldunnplaymint what command shell are you using on windows? ... you might have to use `"` in `cmd` but `'` should be ok in powershell or git-bash

ldunnplaymint:
My bad, looks good

ldunnplaymint (approved):

PR #487, opened 2023-08-01
Fix for under construction building map error

JackPlaymint:
Added fix for construction building ID and missing reference in construction building prefab


chrisfarms (approved):
LGTM .. confirmed I can get under-construction buildling model again without it exploding 😄

PR #507, opened 2023-08-09
cli: unity-bridge

chrisfarms:
### What

* adds `ds -n testnet unity-bridge` command to ds-cli to replace old bridge package
* expose ProjectSettings in Unity to configure which network to point at
* update map to install and run ds-cli from the published npm package when run in editor

### Why

Building the bridge required all the toolchains for building the entire game, but not everyone working on the unity side of things have a full local development environment setup. This change enables pointing the Unity Editor at one of the public deployments of the game, and only has a requirement that `nodejs` and `npm` are installed locally.

### ⚠️ Important changes for map devs

**The ProjectSettings/Downstream settings have changed. You will need to set**:

* a path to the node binary (use `where node` from a terminal to discover path)
* a path to the npx command (use `where npx` from a terminal to discover path)
* the desired network to connect to (use `local` if you using a local `make dev` or Docker deployment, use `testnet` if you want to test against the public deployment or don't have a local dev deployment)
* your private key to test with (don't use one you care about!)

Screenshot 2023-08-09 at 07 54 14

**Occasionally you may need to update the ds cli**

Previously the bridge was always in sync with the current branch you had checked out. This is no longer the case and you may need to update the cli if there have been breaking changes to the bridge.

Update using npm (for people without a dev environment):

```
npm install -g @playmint/ds-cli@latest
```

Update using make (for people with full toolchain):
 
```
make cli
```

... this will overwrite your installed `ds` tool with a manually built one, which is useful for testing changes without publishing the package.

**If you need to change the bridge logic...**

The code that used to be in `./bridge/src/index.ts` --- is now in ---> `./cli/src/commands/bridge.ts`. 

**If you have modified the unity-bridge command, remember to publish npm package**

We don't yet have package publishing automated, so if you make changes to the ds-cli code to change how the bridge works you'll need to run `make publish` by hand to release the package, and then commit back the version bump changes.


hypnoshock (approved):
Great work Farms!

ldunnplaymint (approved):

PR #509, opened 2023-08-09
Stackable buildings

JackPlaymint:
This PR adds a customisable building system that allows you to create a building out of a stack of 2 totems.
There are 9 totems to mix and match.

You can choose what totems to use by modifying the "building" parameter in the deploy script's BuildingUtils.construct function.
The IDs for each totem range from "00" to "08", separated by a hyphen. Here's an example for the cocktail hut:
`BuildingUtils.construct(ds, cocktailHut, "01-00", 5, 13, -18);`

There is a json list that pairs the totem IDs to their respective GameObject names, which can be found in the 'Resources' folder.

Currently, many buildings don't contain totem information, so in order for them not to break the game, any unrecognised totem code will cause the default Tower building to be shown instead of a totem stack.

### DIFF
--- a/map/Assets/Map/Scripts/GameplayElements/MapElementManager.cs
+++ b/map/Assets/Map/Scripts/GameplayElements/MapElementManager.cs
@@ -54,7 +54,11 @@ public class MapElementManager : MonoBehaviour
     {
         Regex rx = new Regex(@"\d{2}");
         MatchCollection matches = rx.Matches(stackCode);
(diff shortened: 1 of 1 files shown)


ldunnplaymint:
Linting :(

hypnoshock (approved):
Looks good. So when we come to implementing the next step of instantiating the extractor building models or the decorative building models, would they have different controllers on them instead of `StackableBuildingController`?

PR #518, opened 2023-08-16
Extraction buildings

hypnoshock:
# What

An addition of `BuildingCategory` on `BuildingKind` which allows us to define if a building is either a blocker, factory, extractor or other.

As part of this I have implemented the extraction category which allows you to define an extractor building that outputs an item that is comprised of one type of goo (atom)

Extractor buildings are show on the map and are the color as defined as their model which is either `red`, `green` or `blue`

As part of the deployment I am deploying 3 extractors that represent the 3 goo types and also placing them on the map close to the starting point.

image

I have also added some basic frontend so the item extraction can be tested

image


ldunnplaymint (approved):

cli/src/utils/manifest.ts
```diff
@@ -109,6 +113,7 @@ export const BuildingKind = z.object({
 
 export const BuildingSpec = z.object({
     name: Name,
+    category: BuildingCategoryEnum,
```
chrisfarms, line 116: This is unused and can be removed I believe (since category is a BuildingKind thing not a Building thing)

contracts/src/fixtures/the-great-cleanup/map.yaml
```diff
@@ -1,656 +1,686 @@
-
 ---
 kind: Building
 spec:
   name: Mecha-Kaiju
-  location: [ 10, -21, 11 ]
+  category: blocker
```
chrisfarms, line 5: category field can be removed from all these Building manifests I believe

cli/src/commands/apply.ts
```diff
@@ -15,16 +22,22 @@ const encodeItemID = ({ name, stackable, goo }: ReturnType<typeof ItemSpec.parse
...
+    const categoryEnum = getBuildingCategoryEnum(category);
+    return solidityPacked(['bytes4', 'uint32', 'uint64', 'uint64'], [NodeSelectors.BuildingKind, 0, id, categoryEnum]);
 };
 
 const encodePluginID = ({ name }) => {
     const id = Number(BigInt.asUintN(32, BigInt(keccak256UTF8(`plugin/${name}`))));
     return CompoundKeyEncoder.encodeUint160(NodeSelectors.ClientPlugin, id);
 };
 
+// TODO: Is there a way of referencing the Solidity enum?
```
chrisfarms, line 36: 🤔 core has the abi (it's how it checks the types for the action strings) ... so can probably expose a category type from there .... but think that one for later

contracts/src/fixtures/extractors/map.yaml
```diff
@@ -0,0 +1,20 @@
+---
+kind: Building
+spec:
+  name: Green Goo Extractor
+  category: extractor
```
chrisfarms, line 5: unused category fields

ldunnplaymint (changes requested):
Sorry, I can't spawn my seeker, keeps hitting this:
[services] 8:23AM ERR subscribe-fail error="context canceled" component=eventwatcher service=indexer simulated=true

ldunnplaymint (approved):
Sorry, I was wrong...

contracts/src/schema/Schema.sol
```diff
@@ -34,8 +34,11 @@ interface Kind {
     function Item() external;
     function CombatSession() external;
     function Hash() external;
+    function BlockNum() external;
 }
 
+uint64 constant BLOCK_TIME_SECS = 10;
```
chrisfarms, line 40: we might need this to be configurable at deployment time ... no change required now

contracts/src/schema/Schema.sol
```diff
@@ -352,4 +384,49 @@ library Schema {
...
+    }
+
+    function setBlockNum(State state, bytes24 kind, uint8 slot, uint64 blockNum) internal {
+        // TODO: don't use generic `Has` selector as it could conflict with something else
+        return state.set(Rel.Has.selector, slot, kind, Node.BlockNum(), blockNum);
+    }
+
+    function getBlockNum(State state, bytes24 kind, uint8 slot) internal view returns (uint64 blockNum) {
+        ( /*bytes24 item*/ , blockNum) = state.get(Rel.Has.selector, slot, kind);
+    }
```
chrisfarms, line 431: think this relationship should probably be something more specific to extraction ... Block can be a generic concept, but think it might be clearer if the relationshpi and these funcs are about the extraction not the blocknum 

```
Node.Something --Rel.LastExtractedAt--> Node.Block
```

contracts/test/rules/BuildingRule.t.sol
```diff
@@ -24,15 +24,38 @@ contract BuildingRuleTest is Test, GameTest {
...
     function testConstructBuilding() public {
         // register a building kind
         bytes24 buildingKind = Node.BuildingKind(20);
         string memory buildingName = "hut";
-        vm.expectEmit(true, true, true, true, address(state));
-        emit AnnotationSet(buildingKind, AnnotationKind.CALLDATA, "name", keccak256(bytes(buildingName)), buildingName);
+        bytes24[4] memory inputItemIDs;
+        uint64[4] memory inputQtys;
+
+        // TODO: Is this test for name annotions necessary anymore?
```
chrisfarms, line 41: no I don't think so ... burn it

Issue #497, opened 2023-08-03
Deploy an interesting set of ~2k Tiles (aka: revealed island shape)


ldunnplaymint:
* use the `/tile-fabricator` to paint a fun shape of Tiles + starting buildings
* export the manifest yaml
* place the manifest yaml into the contracts/src/fixtures dir

PR #526, opened 2023-08-21
tile-fabricator: tool for generating Tile/Building manifests

chrisfarms:
## What

* adds "tile-fabricator": a tool that lets you "paint" tiles / buildings to draw an initial map shape and output the manifests
* fixes a map bug where tiles could not move from discovered -> undiscovered states
* fixes some issues with tsconfig/eslint inconsistencies fighting each other ... and lints a bunch of stuff
* this is all prep to allow someone to do #497 

## Screenshot

Screenshot 2023-08-19 at 20 03 10

_it's quite relaxing to just draw hexagon patterns, like a zen garden_ 

## Usage

It is built into the client frontend ... visit `/tile-fabricator` to load it

Initially you can choose between a `discovered` and `undiscovered` "brush" to paint with (left-click).

Orbit camera with right-click drag

**click "import manifest" to import exisiting Tiles and BuildingKinds** ...for examaple import all the manifests in `contracts/src/fixtures/the-great-cleanup` to get paul's map

Once you have imported other BuildingKinds ... you can "paint" with them to, allowing you to place buildings/ememies/etc on the map.

Toggle the "preview" mode to render the map with the UnityMap renderer to get a better feel for what you have created.

Toggle the "labels" option to show/hide the BuildingKind labels

Click "Export" to download the manifest YAML (which can be placed directly into fixtures)

## Known issues

**The :fren: bug will appear in preview mode if you switch out one building for another** ... this is a unity-map problem where the map fails to handle the case of on an update that changes the `building` field....  I'm not gonna fix this here, but at least now we can recreate it easily 😅 

Screenshot 2023-08-19 at 19 32 48

**map preview crashes if diameter set to larger than ~40** ... I've capped the diameter slider to 40 as workaround which should be enough for now anyway

**loose data on refresh** ... not really a bug, just a missing feature I expect will be desirable as you might spend a while designing a map and then loose it all on one accidental refresh

### DIFF
--- a/cli/src/utils/manifest.ts
+++ b/cli/src/utils/manifest.ts
@@ -189,8 +189,7 @@ const readStdin = async () => {
     return Buffer.concat(chunks).toString('utf8');
 };
 
-export const readManifestsDocumentsSync = async (filename: string): Promise<z.infer<typeof ManifestDocument>[]> => {
-    const filedata = filename === '-' ? await readStdin() : fs.readFileSync(filename).toString();
+export const parseManifestDocuments = (filedata: string, filename?: string): z.infer<typeof ManifestDocument>[] => {
     return YAML.parseAllDocuments(filedata)
         .map((content) => content.toJSON())
         .map((content: any) => {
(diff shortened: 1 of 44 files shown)


ldunnplaymint (approved):

5p0rt5BEArD:
This looks amazing

PR #531, opened 2023-08-21
Removal of category from Building spec

hypnoshock:
#What

Removal of the category field from the Building spec. We are able to look up the BuildingKind to get at the ID so no need to specify the category when defining a building

#Why

Having to define the category when describing an instance of a building was superfluous


ldunnplaymint (approved):

chrisfarms (approved):
👍 

you wouldn't know it existed cos it only appeared two seconds ago ... but could you remove the reference in `frontend/src/pages/tile-fabricator.tsx:264` so that the exported Building manifests dont get a category set 🙏

hypnoshock:
> 👍
> 
> you wouldn't know it existed cos it only appeared two seconds ago ... but could you remove the reference in `frontend/src/pages/tile-fabricator.tsx:264` so that the exported Building manifests dont get a category set 🙏

Sorted ;-)

Issue #528, opened 2023-08-21
dispatch errors not getting returned to player


chrisfarms:
recent changes to batching in services to enable the L2 are preventing revert errors from getting returned to the player

PR #536, opened 2023-08-22
disable simulation node

chrisfarms:
## what

* pins to version of cog that doesn't use action simulation
* updates deployment config to disable simulation
* brings local dev inline with 2s block time so it is representative of testnet
* fixes issues with `ds apply` so that it correctly applies manifests in order without --slow or --batch=1
* sends separate calls of `player.dispatch(...)` in parallel to avoid blocking (await still waits for commit as before)

## why

* The simulation feature was designed to work with a slower block time than the bundling time (ie to make a 10-15s commit time feel like a 1s commit time). We are now running on a 2s commit time, and the simulation stuff as currently designed is not playing nicely and making things unstable.
* There appears to be an issue where the websocket connection fails frequently when forked off the L2 chain... reconnecting and recovering from this failure takes extra time and when occurs means everything stalls to stay in sync ... this is probably a result of network connectivity to the L2 node, ideally we would run a follower node in the same place as the indexer... but that is out of scope so this is quite a hard blocker on fixing this right now

## impact

* 😞 Things are gonna feel a little slower again (response time in the ~2-4s range rather than ~1-2s)
* 😃 Things should be stable again
* 😃 That annoyingly long wait after login is gone
* 😃 fixes: #528

### DIFF
--- a/.devstartup.js
+++ b/.devstartup.js
@@ -15,20 +15,14 @@ const DEPLOYER_PRIVATE_KEY = "0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff
 const commands = [
     {
         name: 'networks',
-        command: "anvil -m 'thunder road vendor cradle rigid subway isolate ridge feel illegal whale lens' --code-size-limit 9999999999999 --gas-limit 9999999999999999 --silent",
-        prefixColor: 'black',
-    },
-
-    {
-        name: 'sim',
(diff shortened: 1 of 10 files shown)


ldunnplaymint (approved):
Moved around, done all the actions, deployed a building. Doesn't feel too bad as it is! Approved.

PR #540, opened 2023-08-22
Only the owner of the extractor is allowed to extract goo from it

hypnoshock:
#What

Only the player that built the extractor can extract goo from it. This is both enforced at the frontend level where a message is displayed if the a player that doesn't own the building tries to use it and also at a rules level. This only applies to the extractors that were built by a player, extractors that are spawned via the map.yaml are public.

On a technical level it's actually the owner of the **output bag** that determines who can use the building or not. Done this way so that we can still have public extractors if we choose to. Also after an extractor has been built, it would be possible for a player to set the output bag to public if they so wished.

I'm also spawning mobile units with 100 of each resource so they can build an extractor (as per sprint goal)

image


contracts/src/rules/BuildingRule.sol
```diff
@@ -251,6 +251,8 @@ contract BuildingRule is Rule {
         if (category == BuildingCategory.EXTRACTOR) {
             // set initial extraction timestamp
             state.setBlockNum(buildingInstance, 0, ctx.clock);
+            // Set output bag owner to player so that only they can take the extracted items
+            state.setOwner(outputBag, Node.Player(ctx.sender));
```
chrisfarms, line 255: Not a huge fan of making the entire category of extractors special here.... but props for finding a super easy way to get the desired result 

... it will mean we have to document this unique behaviour for extractors vs factories ... and it doesn't feel like there would be a compelling rationale for them to behave differently from each other.

could get the same basic end result from the extractor's implementation code without needing to make extractor bags special by creating two different sets of extractors? ... ie public ones and private ones ... and the private ones can simply valididate if you are allowed to press the button

or maybe we just ditch the idea of public ones for now entirely (since we have units spawned with starter stuff anyway) 🤷

chrisfarms:
> When trying to move items from someone else's extractor, I hit an error:

this is not related to this PR ... the revert is expected NoTransferNotYourBag ... the fact you see it caught by the dev/debug handler is an issue I introduced yesterday... that debug handler only catches stuff in local dev not prod builds

ldunnplaymint (approved):
In that case, it is doing what I'd expect

ldunnplaymint (commented):
Merged Farm's fix for the frontend, going to retest it all together

ldunnplaymint (approved):

contracts/src/rules/BuildingRule.sol
hypnoshock: Another reason for me picking the bag way of doing it is that we are going to add permissions to bags eventually which means we'd be able to set the permissions on who else can extract from the extractor other than the owner.

Is the plan for the future to have both factories and extractors use ephemeral bags for output? We still have the problem on factories that you can pinch the things that someone else has just crafted

chrisfarms: I don't believe we are doing "permissions" in the future. The solution was to make it possible that each player is seeing there _own_ output bag, not a special permissioned one. (ie there is never the ability to see something that you cannot take, you just wouldn't see the output)

So the plan is that (once we have the custom item slots ui), you will be able to decide in plugin code to show a player specific bag ... and the implementation can decide to place the output into that bag.

So to me it feels like "stop people taking things from other bags" is a seperate issue from "stop people using a private extractor" ... we can solve the latter using the existing stuff (just revert on use() if you don't own it) .... and the former is problem that we should resolve for ALL building categories not just extractors

don't wanna hold up the PR, since it def does the job, just concerned it violates the princicle of least surprize having extractor outputs work differently than factory outputs

PR #557, opened 2023-08-25
Map blockers

JackPlaymint:
This PR adds decorative blockers to the map.
They're basically buildings that you can't move to.

There's a loverly selection of them ready to go in contracts/src/fixtures/blockers.yaml
Currently you need 25 of each goo to make one, but this should be changed to whatever that recipe should be.

I've also not placed any on the map at the moment, so someone will need to add them in a nice level design-y sort of way.


chrisfarms (approved):
code review LGTM ✅ ... well done finding the manifest spec in cli for adding the model validation... forgot to mention to you that exists!

not spun it up so prob wait for a @ldunnplaymint tick

PR #565, opened 2023-08-29
Doubled camera zoom distance

JackPlaymint:
Changed the maximum zoom distance from 10 to 20.
Obviously this means we're rendering more on screen at a time so will be interesting to see whether this has a large impact on performance.


chrisfarms (approved):
I would approve twice if I could ... zooomies!

PR #572, opened 2023-08-29
map decorations

chrisfarms:
* tom decorated the map and copied some yaml
* I pasted some yaml

### DIFF
--- a/contracts/src/fixtures/map.yaml
+++ b/contracts/src/fixtures/map.yaml
@@ -2,56 +2,56 @@
 ---
 kind: Tile
 spec:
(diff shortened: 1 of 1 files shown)


chrisfarms:
this has been an issue for a while ... it occurs whenever there's a large number of draw calls ... it's not related to this PR, it occurs as soon as there is enough stuff on the map.

I found there's a sorting group on TileContainer that is responsible and it doesn't actually appear to do anything useful ... probably needs a seperate issue raised

ldunnplaymint (approved):

PR #585, opened 2023-08-30
Remove goo under buildings, stop goo z-fighting

JackPlaymint:
This PR hides the goo if any type of building is on top of it (including enemies, blockers and under-construction towers).
If the building is destroyed, the goo reappears.

Also in this PR is a fix for the z-fighting issues that the occur with the goo and the various UI highlights.


ldunnplaymint (approved):
I think we should still show goo under buildings but I think we need a proper solution. I also am not a fan of hiding the goo when you do a move/build highlight but I understand why you chose to do it this way.

PR #592, opened 2023-08-31
configurable max connections for ds apply

chrisfarms:
stop DOSing the chain on deploy 😅 ... makes deploys really slow


chrisfarms:
closing as work in #672 has made it redundant

Issue #344, opened 2023-06-21
The building site should remember the type of building it is going to be


PlaymintPaul:
At the moment, if you click "Build" and click on a building site you have to select the correct building type to see the input materials.

chrisfarms:
was discussing this with baker a while back ... we both agree ... but it will mean we basically need to split construction into START_CONSTRUCTION and END_CONSTRUCTION .... which actually goes well with the ideas of construction taking time so probably isn't the end of the world

PlaymintDavid:
I don't follow. Are you saying it's not possible or not desirable to help a player find out what building type it's going to be?

chrisfarms:
it's both possible and desirable.

but it requires changing how construction works a bit... since it's not just a client-side thing

we are planning to do this. and the end result will be that once construction has started, everyone will know what building it is going to be, and the constructor will not "lose" which building they are mid constructing

PlaymintDavid:
Okay great :)

chrisfarms:
@PlaymintDavid  You'll be happy to hear that it looks like I introduced a bug this sprint that requires this ticket resolved 😅.... so this just task just become high priority and will be sorted soon

@ldunnplaymint ... there's an issue where the building selection disappears during adding the construction materials ... rather than work around it, I'm bumping this as a high priority task to sort tomorrow

ldunnplaymint:
Sounds good to me

PR #597, opened 2023-09-01
one-click construct

chrisfarms:
## what

* makes "construct" a one-click action that works out how to pay for the materials from what is in the unit's inventory
* closes: #344 by making the situation impossible
* resolves the bug where the construct control would disapear midday through your attempt to construct if you used up the last material
* shows the description + cost in words for the selected thing

## note

this means there is no UI to contribute to a "building site"... but since this "unknowable partial construction" problem has come up a few times and causes confusion, this was agreed with paul as a quick fix for the issue.

it also feels less tedious to make a building this way so maybe it's preferable long term anyway 🤷 

Screenshot 2023-09-01 at 07 19 51

### DIFF
--- a/frontend/src/components/organisms/logs/logs.styles.ts
+++ b/frontend/src/components/organisms/logs/logs.styles.ts
@@ -10,7 +10,7 @@ import { LogsProps } from './index';
  * @return Base styles for the tile action component
  */
 const baseStyles = (_: Partial<LogsProps>) => css`
-    max-width: 80rem;
+    max-width: 33vw;
     max-height: 20rem;
 
     > .log {
--- a/frontend/src/components/views/shell/shell.styles.ts
+++ b/frontend/src/components/views/shell/shell.styles.ts
@@ -15,11 +15,6 @@ const baseStyles = (_: Partial<ShellProps>) => css`
     display: flex;
     flex-direction: column;
 
-    > .controls {
-        width: 33vw;
-        margin-bottom: 2rem;
-    }
-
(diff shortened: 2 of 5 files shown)


ldunnplaymint:
This feels abit more confusing to me. I get that the buildings go into the unconstrucable pile when I cannot make them but 1) It is now loads of effort to find out what buildings need to be constructed and 2) Having the unconstrucable pile also contain the "1337 br1c|<z" buildings makes it quite unwieldy. I also imagine that list will just be horrendous when other people start deploying.

chrisfarms:
I haven't changed the behaviour of the unconstructable pile here?

ldunnplaymint:
I mean that when I cannot afford to construct a building, it gets added to that pile. Before, it was basically the "I can ignore this" pile in my head, now I have to sift through it if I want to remember what a construction costs.

chrisfarms:
well yes ... it was broken before and should not have been showing you something you cannot build 😅 

we have not got a good solution for this selection yet... the previously methods have not worked and this isn't great either ... but the main things we are trying to archive are:

* Your own deployed buildings are always at the top
* It groups things to help you a bit
* try not to overwealm someone with a list of crap they can't build (unless they opt in)

Your desire seems to be "I want to find out what buildings I might be able to make in the future" ... which is a totally legit desire, and something that we should have a solution for ... but not something this selection list is ever going to be good for, needs a proper library of all the things you can filter and search ... not a picker

ldunnplaymint (approved):

PR #652, opened 2023-09-06
Added docs for building totem IDs

JackPlaymint:
This PR adds a page to the Downstream docs which explains how to use totem IDs for customising building visuals.
It also provides a visual list of the IDs and their models.

### DIFF
--- a/docs/content/code-docs/building-model-ids.md
+++ b/docs/content/code-docs/building-model-ids.md
@@ -1,0 +1,28 @@
+# Building Model IDs
+
+Building visuals are constructed using two stacked 'totems'.
+You can choose which totems to use for your building by setting the `model` tag in your buildings manifest.
+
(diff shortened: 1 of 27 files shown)


chrisfarms (commented):
noice 👍 


Should we also call out the differences between the model options available dependent on what building `category` is set?

ie.

when `category` is `"blocker"` then options for model are:

    'enemy',
    'CactusLarge',
    'CactusSmall',
    'GrassLarge',
    'LogWall',
    'OakTreesLarge',
    'OakTreesSmall',
    'PalmTrees',
    'PineTreesLarge',
    'PineTreesSmall',
    'rocksLarge',
    'rocksSmall',
    'Shrub',
    'StoneWall',

when `category` is `"factory"` or `"custom"` then options for `model` are:

   -- totems stuff from this PR

when `category` is `"extractor"` then options for `model` are:

red, green, blue


maybe this is too much scope creep ... maybe the minimum is to mention these totem ids are only valid when `category` is `factory` or `custom`

chrisfarms (approved):
ledge!

PR #665, opened 2023-09-08
Slowed down goo rates, and changed tiles to display goo per hour

PlaymintPaul:
- Goo extraction is about 5x slower on the best tiles (slightly better on the worst tiles)
- Goo rate on tile display is now shown in goo per hour
- Edited Generic Extractor.js to deal with slow goo per seconds (rather than rounding to 0)


chrisfarms:
> Trying to work out why the test is failing...

think you need to increase the `vm.roll()` call in the `testExtraction` test amount in the test ... since you slowed down extraction

vm.roll is like fast forwarding to the future

chrisfarms (approved):

PR #671, opened 2023-09-13
Selecting an invalid tile in move Intent does NOT cancel move intent

### LINKED ISSUE #615: Multiclick move sometimes resets
https://github.com/playmint/ds/assets/129759334/e9c52b41-2d33-4b3a-bbdd-50e3ee6f5997

When I multiclick a long path sometimes it resets, meaning I have to draw my path all over again. It's so annoying!

### LINKED ISSUE #636: Only allow movement to be cancelled via the cancel button
Remove the ability to click the map to cancel movement

5p0rt5BEArD:
### What
The move intent, as handled by the map, no longer cancels if the player clicks an invalid tile. Instead the current path does not change and the player remains in move intent.

To cancel move the player must not click the UI cancel cross, or click an empty space with no tiles.

### Why
Its too easy to cancel by mistake

fixes #615 
fixes #636

### DIFF
--- a/map/Assets/Map/Scripts/Intent/MoveIntent.cs
+++ b/map/Assets/Map/Scripts/Intent/MoveIntent.cs
@@ -338,7 +338,12 @@ public class MoveIntent : IntentHandler
             GameStateMediator.Instance.SendSelectTileMsg(tileIDs);
         }
         else
-            DeselectMobileUnitAndIntent(true);
+        {
+            // invalid selection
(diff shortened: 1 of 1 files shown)


ldunnplaymint:
We want to run a build off with this change so we have something to playtest over the next week or so. Can we get it in just for that and we don't care if it gets gutted later?

ldunnplaymint (approved):

PR #685, opened 2023-09-27
feat: allow content-type headers in the fetch request

goblinoats:


core/src/plugins.ts
```diff
@@ -462,6 +462,13 @@ export function loadPlugin(runtime: QuickJSRuntime, dispatch: DispatchFunc, logg
         if (unsafeOpts.body) {
             allowedOpts.body = unsafeOpts.body;
         }
+        if (unsafeOpts.headers) {
+            if (!!unsafeOpts.headers['Content-Type']) {
```
chrisfarms, line 466: can we do a `typeof` check for string here rather than truthyness since this is coming from untrusted plugin land 🙏

ldunnplaymint (approved):

PR #686, opened 2023-09-27
rendering the docs markdown from frontend

chrisfarms:
### what

* adds a `/docs/*` Page that (rather shodidly) renders the docs/*.md files.
* adds a (rather shoddy) docs/map buttons to nav bar
* strips out the old docs site
* made a little sidebar that is staticly generated from the docs/ folder tree structure to get us started

### why

* so we can integrate the map screen and additional content screens 
* so we have less stuff to maintain, build, deploy
* so we can style things more centrally

### todo

I haven't really styled anything other than making the text white and picking a code highlight theme that is blue 😅 

... so it looks like 

Screenshot 2023-09-27 at 15 05 17


ldunnplaymint (approved):

ldunnplaymint:
Looks good but we need some loading states. Switching from Map to Docs takes a few seconds and there are no on screen indicators the game is doing anything.

PR #697, opened 2023-10-04
Remove mermaid so we can also stop using Playwright

ldunnplaymint:


hypnoshock (approved):
Approved so it's unblocked however we might want to remove playwright from the `devDependencies` in `package.json` if it's not being used anymore

frontend/package.json
```diff
@@ -4,9 +4,9 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "dev": "playwright install --with-deps chromium && next dev",
-    "build": "playwright install --with-deps chromium && next build",
-    "export": "playwright install --with-deps chromium && next build && next export",
```
hypnoshock, line 9: line 74 still has playwright as a dev dependency. I suppose we can get rid of it...

ldunnplaymint: Done

PR #711, opened 2023-10-12
Quest system

hypnoshock:
# What

- A basic quest system where data is represented as `Quest` and `Task` nodes.
- Quests along with each of their tasks are defined in [yaml](https://github.com/playmint/ds/blob/91e6fe9d09448c019098f92b2d03289cb33bd138/contracts/src/fixtures/quests/intro.yaml) using our current manifest system
- Tasks come in different kinds which have different completion criteria for example the `location` kind of task evaluates completion by checking the player's unit location against the location associated with the task.
- Quests can be linked together so that when a quest is completed, a set of up to 5 quests can be auto accepted
- Accepted quests are displayed in a quest panel with their associated tasks displaying their current completion state
- When the frontend determines a quest as complete, a button will appear to 'complete quest' which will validate the task completion on chain and if satisfied will set the quest status to `completed`
	- Quest status is stored as the weight on the `hasQuest` edge which connects a `Player` node to a `Quest` node
	    - `NONE = 0`, `ACCEPTED = 1`, `COMPLETED = 2` 

## The panel
The panel supports multiple quests with a crude accordion implementation
image

### DIFF
--- a/cli/src/commands/apply.ts
+++ b/cli/src/commands/apply.ts
@@ -269,16 +269,27 @@ const questDeploymentActions = async (
     const spec = file.manifest.spec;
     const existingItems = await getManifestsByKind(ctx, ['Item']);
 
+    const existingBuildingKinds = await getManifestsByKind(ctx, ['BuildingKind']);
+    const pendingBuildingKinds = files.map((doc) => doc.manifest).filter(({ kind }) => kind === 'BuildingKind');
+
     const encodeTaskData = (task: (typeof spec.tasks)[0]) => {
+        const coder = AbiCoder.defaultAbiCoder();
+
         switch (task.kind) {
             case 'coord': {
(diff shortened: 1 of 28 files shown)


ldunnplaymint:
While I don't think this is your branchspecifically, having a building selected while a unit is moving causes a lot of rapid stuttering on the 3D model moving from tile to tile. I can't see this on main.

hypnoshock:
> Looks like some stuttering has been introduced in your branch. If you move a unit and then have a tile/building selected and the info panel open, the unit will stutter as it moves tile to tile. 

I'm seeing the same stuttering in main if I build a cocktail hut and craft a cocktail. I think the problem has something to do with item kinds which isn't localised to my branch. Please may you retest and let me know if you cannot replicate in main.

ldunnplaymint (approved):

PR #714, opened 2023-10-13
A much more cut down map

ldunnplaymint:
Intention is to remove as much as possible to stop Main from falling over.


PlaymintPaul:
I added a handful of tiles to link up some buildings.
Removed a few buildings that were too adrift to bother linking up
Renamed the files in temp_fixtures so that we know where they belong

ldunnplaymint:
Cheers, assuming it deploys and the tests finish, I will merge it

PR #738, opened 2023-10-20
player preference to pick render quality

chrisfarms:
allow player to pick from a few fixed render qualities (480p, 720p 1080p) as well as "auto" (which takes the browser window size) and "native" (which takes the browser window size + packing in more pixels for hiDPI screens)

preference is persisted in localStorage

default is "auto" ... and for retina screens I think auto is going to lower the resolution from the previous default .... change to "native" if you want your full hiDPI res back

the control to set the quality is hidden in the "account" dialog (click the player icon top left in nav bar).... longer term we will move this out into a "settings" dialog or something ... but this is just a quick starting point

Screenshot 2023-10-20 at 08 43 46

zooming out should no longer cause it to attempt to render a crazy high res ... zooming in however _will_ make the res worse, which is weird, but since zooming in will prettymuch break the rest of the UI anyway, I think we can live with that quirk

### DIFF
--- a/frontend/src/components/panels/nav-panel.tsx
+++ b/frontend/src/components/panels/nav-panel.tsx
@@ -7,6 +7,9 @@ import styled from 'styled-components';
 import { Dialog } from '../molecules/dialog';
 import Link from 'next/link';
 import { useRouter } from 'next/router';
+import { GlobalUnityContext } from '@app/hooks/use-unity-instance';
+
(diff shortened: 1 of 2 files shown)


ldunnplaymint (approved):

ldunnplaymint:
Works and what not, no errors or obvious problems. The major issue is the unit highlight looking very chonky on lower resolutions but that might just be what it is like running on a high rest screen. Might be worth a Jack pass at a later date?

PR #762, opened 2023-10-25
Apply color to buildings on map

JackPlaymint:
It's a bit of a work around due to variable limitations.
First we take the color from the yaml and combine it into the model variable's string.
Then we decode that string back into separate model and color variables before sending it back to Unity to update the building model color.


chrisfarms (approved):

hypnoshock (approved):
Solidity stack limit! >:-(
The workaround LGTM

PR #764, opened 2023-10-25
Corrupted User Quest chain improvements

PlaymintPaul:
- Easier to complete
- Fixed an issue with the World Constructor
- Added reactors for the other color goos


ldunnplaymint:
I think the corrupted user quest should direct you towards the industrial espionage building in the text. I first looked and found the corrupted user and decided to gather the goo it wanted only to still be confused. After clicking around I realised I should have gone to the spy store first. 

Gonna get this merged now regardless for tomorrow's playtest.

ldunnplaymint (approved):

PlaymintPaul:
> I think the corrupted user quest should direct you towards the industrial espionage building in the text. I first looked and found the corrupted user and decided to gather the goo it wanted only to still be confused. After clicking around I realised I should have gone to the spy store first.

It does.  The quest has 1 task which is to get a disguise,  and the Eye button focuses on the disguise shop

ldunnplaymint:
The quest panel I was shown is here:
Screenshot 2023-10-26 at 08 07 04


The text didn't feel clear to me I needed to go to the espionage building to work out I need a shirt and tie. I'd rather it was more explicit about what I needed. I.E. "You will need a disguise" or "I hear there is a Spy shop to the east" or "Mask you Identity with a Boring Disguise". Atm I have to deduce that the game wants that specific item.

Also, I don't think we should rely on the button to highlight progression.

PlaymintPaul:
I guess the text can be tightened a bit

PR #785, opened 2023-10-27
Added custom tooltip element

JackPlaymint:
This PR replaces the default browser based tooltip with a custom element.
This gives us more control over the look and feel of the tooltip, while also allowing us to have it appear faster than the browser based one.


ldunnplaymint:
The tool tip can go off screen if it is too big
Screenshot 2023-10-30 at 08 18 17

ldunnplaymint:
Empty slots are using the old tool tip

Screenshot 2023-10-30 at 08 28 17

ldunnplaymint (approved):

PR #786, opened 2023-10-27
Extractor speed up

PlaymintPaul:
- Made extractors cheaper to build
- Goo rate is 5x the speed
- Placed some extractors on the map at deployment time


hypnoshock (approved):
LGTM

Issue #771, opened 2023-10-26
end combat without opening modal


ChrisWh99:
It would be nice to be able to end combat without having to go into the View window

PR #788, opened 2023-10-27
If combat has ended it is now possible to end combat from the summary

hypnoshock:
# What

image

# Why

Going into the modal to end combat was long winded

Solves: [#771](https://github.com/playmint/ds/issues/771)


chrisfarms (approved):
now i can deforest the land faster

PR #797, opened 2023-10-30
Item names must now be unique when registering an item.

hypnoshock:
# What

Item names must now be unique when registering an item. Item node IDs are formed from the item's name AND composition which meant it was possible to register two items with the same name providing they had a different composition (this was actually a feature, not a bug!). To enforce unique names across items we now attach an ID node to an item whose owner is the item node. This allows us to check if any given ID is already 'owned' by a node.

# Why not make the Item node IDs the name / hash of the name?

Because we use the Item node ID as a static store of data i.e. we decode the ID to find out the atom values of a item. If we were to change what constitutes an item ID we'd need to store the item's composition as nodes and refactor a fair amount of code which I believe is a significant risk.

# Why

We treat name annotations as if they are unique but this is not currently enforced on the contract side which causes problems when you do manage to hit a conflict

# How to test

Register a Squircle with a different composition from the one deployed as part of the fixtures:

```
---
kind: Item
spec:
  name: Squircle
  icon: 31-286
  contract:
    file: ./Squircle.sol
  goo:
    red: 11
    green: 10
    blue: 10
  stackable: true
```

Expect:

```
1 manifests failed to apply:

file: squircle.yaml
error:Error: failed commit batch tx: execution reverted: item name already registered

```

### DIFF
--- a/contracts/src/Downstream.sol
+++ b/contracts/src/Downstream.sol
@@ -67,6 +67,7 @@ contract DownstreamGame is BaseGame {
         state.registerNodeType(Kind.BlockNum.selector, "BlockNum", CompoundKeyKind.UINT160);
         state.registerNodeType(Kind.Quest.selector, "Quest", CompoundKeyKind.UINT160);
         state.registerNodeType(Kind.Task.selector, "Task", CompoundKeyKind.UINT32_ARRAY);
+        state.registerNodeType(Kind.ID.selector, "ID", CompoundKeyKind.BYTES);
 
         // register the relationship ids we are using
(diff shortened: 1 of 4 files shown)


ldunnplaymint (approved):

chrisfarms (approved):
yep, suitably horrible 😅  ... we need a better system for these annotations since they aren't really annotations at this point heh

I think we might need to do the same for building name annotations too 🤔  (similar issues when writing manifests that only use names not ids) .... could go in this PR if easy or could be a follow up, don't think there's an issue explicitly for buildingkind names

Issue #780, opened 2023-10-26
MVP building builder for Iain's workshop


ldunnplaymint:
Enough to build the Squircle Factory and the Very Important Thing factory, then export .sol, .json and .yaml files to deploy using the CLI.

PR #798, opened 2023-10-30
MVP building fabricator

chrisfarms:
### what

This is an MVP of a tool to quickly generate/apply a basic factory type building to the game.

Screenshot 2023-10-30 at 14 07 23

Screenshot 2023-10-30 at 14 07 31

Screenshot 2023-10-30 at 14 07 38

* You can find it at `/building-fabricator`
* "Apply" will deploy directly to the game if you are connected
* "Export" will download a zip containing the yaml + sol + js for your factory
* Validation is poor to non-existant at this stage, but if you fill everything in it should work
* Had to rejig the cli bits so I can import the funcs to apply - which creates some tech debt to move things from cli -> core since they are now shared
* It's very much an "inspired by" the design, not the actual design at this point ... but this gets us started
* does some weird stuff to get unity to be sized correctly
* fixes a few unity bits to get models to update

resolves: #780


ldunnplaymint:
Seeing a lot of errors firing (
[errors.txt](https://github.com/playmint/ds/files/13206544/errors.txt)
)

Position of the 3D viewer seems to stay in it's position when resizing the screen.

I can see the unit scene loading bar in the background

The Unity scene is visible before the Hexwood style UI when loading in

Selecting an existing item doesn't work
Screenshot 2023-10-30 at 15 06 12

chrisfarms:
Can you expand on "Selecting an existing item doesn't work" 🙏

5p0rt5BEArD:
Deploying via the cli seems broken to me.

I get two "file not found" errors when running `node cli/dist/ds.js apply -n 'local' -f -R contracts/src/example-plugins/SquircleFactory/SquircleFactoryManifest.yaml`

ldunnplaymint:
> Can you expand on "Selecting an existing item doesn't work" 🙏

My bad. I cannot select some options in the drop down for using an existing item as an output but I am not told why.

The 3 good, very important stuff, gold note... just to name a few.

5p0rt5BEArD:
I think it makes more sense for the output tab to appear to the left if the input

chrisfarms:
> The 3 good, very important stuff, gold note... just to name a few.

gotcha ... i added some validation and all our items are invalid heh ... willfix

chrisfarms:
> Seeing a lot of errors firing ([errors.txt](https://github.com/playmint/ds/files/13206544/errors.txt))

resolved.

> Position of the 3D viewer seems to stay in it's position when resizing the screen.

improved


> I can see the unit scene loading bar in the background

disabled the loading on this page, so now you just have to wait for it to appear ... will prob need to rewrite the loading stuff if we want it not fullscreen

> The Unity scene is visible before the Hexwood style UI when loading in

improved, same as above, will just appear when ready

> Selecting an existing item doesn't work

resolved ... spaces now allowed in names

chrisfarms:
> Deploying via the cli seems broken to me.
> 
> I get two "file not found" errors when running `node cli/dist/ds.js apply -n 'local' -f -R contracts/src/example-plugins/SquircleFactory/SquircleFactoryManifest.yaml`

Looks like you have your command flags mixed up. I think you want:

```
ds apply -n local -R -f contracts/src/example-plugins/SquircleFactory/
```

.. for applying a directory... or.... 

```
ds apply -n local -f contracts/src/example-plugins/SquircleFactory/SquircleFactoryManifest.yaml
```

... for a single manifest

ldunnplaymint:
Cannot select "None" after selecting an item from the existing items list

Items in the existing items list are un-useable. I tried to use our Goo and got this:
```failed commit batch tx: execution reverted: you are not the owner of the output item and no item contract exists to ask for permission
```

ldunnplaymint:
Should I be able to set how many of a stackable item to output

ldunnplaymint:
Switching back to the map from the top buttons displays a half screen and broken map. I think we aren't planning on allowing the top nav navigation? We should turn them off on the building-fab page if so.

chrisfarms:
> Cannot select "None" after selecting an item from the existing items list

None isn't an item ... so you can't select it.

> Should I be able to set how many of a stackable item to output?

Not in this version.

> Items in the existing items list are un-useable. 

Yep, that error is expected, the item contract decides if you are allowed to register someone else's item

chrisfarms:
> Switching back to the map from the top buttons displays a half screen and broken map. I think we aren't planning on allowing the top nav navigation? We should turn them off on the building-fab page if so.

those buttons are global nav ... so will disappear when we remove them

ldunnplaymint:
RE: None in the list
Feels like having an list item I cannot select is weird

RE: Expected erroring items
Can we hide items you haven't got access to ask the contract for use of? List is already very full and having options you cannot use feels wrong to me

RE: everything else
Ok :+1:

chrisfarms:
> RE: None in the list Feels like having an list item I cannot select is weird

I've allowed you to select none ... but I think it's weird to let you pick an invalid option

> RE: Expected erroring items Can we hide items you haven't got access to ask the contract for use of? List is already very full and having options you cannot use feels wrong to me

There is no way to know if an item can be used until you attempt to register it ... it's not a flag, it's a function call ... and that function could be just doing Math.random() to decide if you are allowed or not ... so unfortunately not much we can do at this point

ldunnplaymint:
Maybe there is a better way like not having None in the drop down or something? I don't feel THAT strongly about it so basically happy where it is now. Approved.

ldunnplaymint (approved):

Issue #800, opened 2023-10-31
Metamask cant be rejectected - gets stuck in signature request loop


5p0rt5BEArD:
- click connect
- click metamask
- reject signature

at this point nothing will prevent metamask from re-asking for signature including

- clciking x on the wiating to connect dialogue
- refreshing page
- clearing cache with cmd-shift-r

Issue #766, opened 2023-10-25
I cannot scan the wallet connect QR code


ldunnplaymint:
I have tested some other QR codes and they have worked. I think potentially it's because of the new white background.

ldunnplaymint:
Happening again. We should add a black border around the QR code. Here is a mockup:
<img width="452" alt="Screenshot 2023-10-31 at 08 17 51" src="https://github.com/playmint/ds/assets/99176798/e9e0b616-970a-418b-9a16-75691a7755ed">

Issue #752, opened 2023-10-24
option to connect with a burner wallet


chrisfarms:
optionally allow burner/throwaway/ wallets to be used for ephemeral playtesting accounts

should be something we can turn on/off via `frontend/public/config.json`

PR #802, opened 2023-10-31
burner wallets and misc wallet fixes

chrisfarms:
### what

* allow enabling ephemeral "burner" wallets by setting `{wallets: {burner: true}}` in the frontend config.json
* burner key stored in localstorage so will persist past refresh, but destroyed on "disconnect"
* fix localstorage sessions getting wiped out too frequently
* only ever do session loading once
* make the QR render for wallet connect bigger, brighter, and remove the excavated image

### related

* resolves: #752 
* resolves: #766 (hopefully... please test since it always works on my phone)
* resolves: #800

### DIFF
--- a/core/src/wallet.ts
+++ b/core/src/wallet.ts
@@ -3,10 +3,22 @@ import { concat, filter, fromPromise, fromValue, lazy, makeSubject, map, onEnd, 
 import { CompoundKeyEncoder, NodeSelectors } from './helpers';
 import { EthereumProvider, Selector, Wallet } from './types';
 
-export type WalletProvider = {
-    method: string;
-    provider: EthereumProvider;
-};
+export type WalletProvider =
+    | {
+          method: 'walletconnect';
+          provider: EthereumProvider;
+          address?: string;
+      }
+    | {
+          method: 'metamask';
+          provider: EthereumProvider;
+          address?: string;
+      }
+    | {
+          method: 'burner';
+          provider: ethers.Wallet | ethers.HDNodeWallet;
+          address?: string;
+      };
 
 export function makeWallet(): {
     wallet: Source<Wallet | undefined>;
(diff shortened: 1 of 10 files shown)


ldunnplaymint:
If I do the following:

Press connect > Select Meta Mask >Reject Metamask sign request

My metamask wallet address is visible in the top left and clicking it brings up the disconnect dialogue. Obviously I am not connected but I don't think the UI handles that flow.

5p0rt5BEArD:
> If I do the following:
> 
> Press connect > Select Meta Mask > Reject Metamask sign request > Close the waiting dialogue
> 
> My metamask wallet address is visible in the top left and clicking it brings up the disconnect dialogue. Obviously I am not connected but I don't think the UI handles that flow.

This flow was not handled before this pr - I entered an issue just before (issue #800)

5p0rt5BEArD:
Wallet connect works for me with these changes 👍

chrisfarms:
> My metamask wallet address is visible in the top left and clicking it brings up the disconnect dialogue. Obviously I am not connected but I don't think the UI handles that flow.

you _are_ connected, but you don't have a session because you rejected creating the session ... this should be ok, you will get prompted for a session if you try and do anything that needs one.

ldunnplaymint:
> > My metamask wallet address is visible in the top left and clicking it brings up the disconnect dialogue. Obviously I am not connected but I don't think the UI handles that flow.
> 
> you _are_ connected, but you don't have a session because you rejected creating the session ... this should be ok, you will get prompted for a session if you try and do anything that needs one.

If you have never spawned a unit, the only thing you can do is click the WID and disconnect. Feels a bit weird but it's an edge case so I am happy to let it slide... THIS TIME.

5p0rt5BEArD (approved):
Glanced at code
tested walletConnect signing
tested burner wallet sign in 
tested transition between building-fabricator and map
- works with burner with no additional signing
- works with metamask with no additional signing
- requires re-signing with wallet connect (same behaviour as before)
tested rejecting metamask sign and it works (fixes #800)

PR #814, opened 2023-10-31
render docs in orange/black/white

chrisfarms:
### what

some lightweight docs styling

* use black/orange/white colorscheme for docs
* pick a light syntax highlighht theme
* give some active states for the sidebar
* make it fit on screen bit better

...this PR doesn't attempt anything more cleverer than making the things readable again

resolves: #779 

Screenshot 2023-10-31 at 14 36 05


ldunnplaymint:
I was just about to complain about this lol

ldunnplaymint:
I don't know if this is a bug or just weird functionality but the scroll position persists between pages.

Scroll to bottom of page 1, switch to page 2, note scroll position

chrisfarms:
> Do we need to kill all the mermaid stuff? Worth doing here (If not already done).

nothing needs doing here (other than removing the docs we don't want) which I'm not doing in this PR as don't want to conflict with any other docs changes

ldunnplaymint (approved):

chrisfarms:
> I don't know if this is a bug or just weird functionality but the scroll position persists between pages.
> 
> Scroll to bottom of page 1, switch to page 2, note scroll position

I think this is only a thing in the dev server not on prod build

Issue #806, opened 2023-10-31
Orientation quest window overlap


PlaymintDavid:
Windows overlap too much on my screen default zoom level. Looks great on 80% zoom.

<img width="1440" alt="Screenshot 2023-10-31 at 09 36 24" src="https://github.com/playmint/ds/assets/129759334/992ea939-78c7-4da9-8b49-4a94b3bd6109">

Issue #777, opened 2023-10-26
Stuff gets pushed off of the screen still


ldunnplaymint:
Make this better inabit

PR #824, opened 2023-11-01
squishing the HUD to better support tiny screens

chrisfarms:
### What

Shrinking, Squishing, and Shifting around HUD elements as things are not playing nicely on smaller screens:

* reduced font-sizes and padding all over the place to claim back pixels
* switched from grid to flex-box to arrange things more flexy
* removed the nav links and claimed the vertical space back for the RHS
* allow the RHS to scroll if it overflows (hacky - but better than being totally unusable)
* rearranged and squished the unit panel to use less vertical space
* the RHS bar will now allow you to scroll it if it overflows, with a fade-out effect to hint that there is more content.

Here's a scenerio of an iPad mini view of the game with too much stuff:

https://github.com/playmint/ds/assets/45921/b801c57c-af1d-4731-a762-0f88e2991363

Here's running with David's unusual aspect radio where all the vertical real estate is taken up by things like docks,bookmarks, apple menu bar etc

https://github.com/playmint/ds/assets/45921/c5ab0692-7765-47b3-9605-4c03076df7be

### Note

Squishing everything and having hidden scrolling areas isn't good! we really need a HUD design that better accommodates all the stuff so this is very much hammering that square peg in the round hole for now.

### Related

* resolves: #777
* resolves: #806 (at least improves it about as far as we can get for now without an overhaul)

### DIFF
--- a/frontend/src/components/organisms/onboarding/index.tsx
+++ b/frontend/src/components/organisms/onboarding/index.tsx
@@ -11,6 +11,7 @@ export interface OnboardingProps {
 }
 
 const StyledOnboarding = styled(StyledHeaderPanel)`
+    width: 40rem;
     h3 {
         margin: 0;
     }
--- a/frontend/src/components/panels/item-plugin-panel.tsx
+++ b/frontend/src/components/panels/item-plugin-panel.tsx
@@ -5,6 +5,7 @@ import { BasePanelStyles } from '@app/styles/base-panel.styles';
 
 const StyledItemPluginPanel = styled.div`
     ${BasePanelStyles}
+    width: 26rem;
 `;
 
 export const ItemPluginPanel = ({ ui }: { ui: PluginUpdateResponse[] }) => {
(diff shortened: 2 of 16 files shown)


ldunnplaymint:
Item plugin is a bit big
Screenshot 2023-11-01 at 09 50 31

ldunnplaymint:
Was this PR meant to fix this? This is at 100% zoom
Screenshot 2023-11-01 at 10 20 23

hypnoshock:
> Was this PR meant to fix this? This is at 100% zoom Screenshot 2023-11-01 at 10 20 23

Going from the description of the PR, it should be scrollable at least. There isn't really any clean 'solution' without a proper design. I think we were talking about being minimise and maximize panels but that would be out of the scope of this PR

chrisfarms:
> Was this PR meant to fix this? This is at 100% zoom Screenshot 2023-11-01 at 10 20 23

define "fix" 😆 

the RHS should now be scrollable when it overflows, and generally font-sizes and padding have been reduced accross the board to give the RHS bit more space.

It's still going to overflow... it's impossible to prevent that in this design as there are multiple variable height content areas on the RHS ... so it's about damage control not prevention

hypnoshock (approved):
LGTM. Doing the best with what we have here, there isn't a solution without a design to accommodate panels of variable height

ldunnplaymint (approved):
I can't see any new problems and it feels a bit better to me. Farm's can decide if he wants to merge!

PR #833, opened 2023-11-01
Progress bar on tasks

hypnoshock:
# What

Quests show progress bar

image
image

# Known bug

If you have partially completed a quest and reload the page, if that quest is minimised, the progress will be at 0% until it is maximised. This is because the tasks are only evaluated at the point they are listed out within a quest.


chrisfarms (approved):
```
[=================] 1/1
```

^--- approval progress bar for you

ldunnplaymint (approved):

Issue #732, opened 2023-10-19
I want to be able to minimise the QUESTS UI


ldunnplaymint:
Can we add a minimise button so I can hide it when I am not actively questing?

PR #834, opened 2023-11-01
toggle button for quests panel

chrisfarms:
### what

* adds the toggle button for quests panel
* makes the player icon more like the figma
* resolves: #732 

Screenshot 2023-11-01 at 13 35 06

### know issue

"fake" tasks (like clicking buttons) get unticked if you toggle the panel, since they don't really exist


hypnoshock (approved):
LGTM. Saw the fake task bug coming 🫤

Issue #828, opened 2023-11-01
Add Combat stats to Unit and Building UI


PlaymintPaul:
See "Combat Stat Display" thread in dev-threads for mockup

PR #837, opened 2023-11-01
add little labels for atk/def/life stats in unit/building panels

chrisfarms:
* resolves: #828


ldunnplaymint:
Functionally fine, will approve but first:

Can we do some formatting to make them more prominent? 

Making the Atk Def Life next bold, colour the text so it obviously corresponds to goo types? 

@PlaymintPaul Any thoughts?

chrisfarms:
> Can we do some formatting to make them more prominent?

there was a discussion a while ago that we don't want them to be seen as a score, so they shouldn't be too prominent ... this "design" is just the thing from the dev-threads to surface them easy to enough to style, but we just don't have a style yet so subtle and hidden :)

ldunnplaymint (approved):

PR #842, opened 2023-11-02
Fabricator exportes js updated

5p0rt5BEArD:
example-plugin BasicFactory js code copied to the The BASIC_FACTORY_JS hardcoded string of code


frontend/src/pages/building-fabricator.tsx
```diff
@@ -250,75 +250,109 @@ contract BasicFactory is BuildingKind {
...
-        if (!selectedUnit) {
-            ds.log("no selected engineer");
-            return;
-        }
-        if (!selectedBuilding) {
-            ds.log("no selected building");
+        const mobileUnit = getMobileUnit(state);
+    
+        if (!mobileUnit) {
+            ds.log('no selected unit');
```
chrisfarms, line 265: we should probably not have `ds.log` calls in this example since it doesn't do anything interesting ... switch to console.log maybe

chrisfarms (approved):
minor comment, non-blocking

frontend/src/pages/building-fabricator.tsx
5p0rt5BEArD: Ah yes - I'll fix that in the actual example code and in this copy in a future pr

PR #849, opened 2023-11-02
Added more protection to accepting quests

hypnoshock:
# What

We now check all quests up until the given questNum to verify that the player hasn't accepted it in the past. This should solve the problem where it was possible to accept a quest twice. It would have been possible to accept a quest more than once providing the questNum (which is the edge) wasn't pointing to a quest as we were only check if the slot was free.

# Why

In play tests we have experienced a couple of times where duplicate quests have been listed.


chrisfarms (approved):
didn't spin it up but LGTM

Issue #794, opened 2023-10-30
Change the target focus QUEST icon for a Crosshair


ldunnplaymint:
A cross hair is less confusing

ldunnplaymint:
Need an icon

chrisfarms:
![Image](https://github.com/playmint/ds/assets/45921/cd8d7a10-4f67-4bbb-9644-7ea22f2ba730)



![Image](https://github.com/playmint/ds/assets/45921/9353aa62-aca8-4683-bff1-0255781d896b)

this one is in the figma

PR #854, opened 2023-11-02
crosshair task location icon

chrisfarms:
* resolves: #794


hypnoshock (approved):
We did this one at the same time. I prefer yours as I'm loading the file in and with the way you're doing it, we can recolour the SVG on `:hover` if we want to wheras loading it in as an `<Image>` like I did won't let you do that

PR #855, opened 2023-11-02
Quest focus button icon changed to the one found in Figma

hypnoshock:
# What

Changed quest focus button to match the one found in Figma

image

resolves: #854


chrisfarms:
heh gotta claim those issues quick ... they are flying off the shelves 😆

Issue #625, opened 2023-09-04
incorrect building images in combat modal


chrisfarms:
still showing castles in combat modal

chrisfarms:
what should we use instead of the castle? ... we don't have pics for every building combination, maybe some kind of pictogram/icon would be better here

PR #857, opened 2023-11-02
remove icons from combat modal

chrisfarms:
* removes participant icons from combat modal
* resolves: #625


5p0rt5BEArD (approved):
LGTM and tested

Issue #847, opened 2023-11-02
something catching the clicks on unit icon to focus camera


chrisfarms:
weirdly hard to hit click-area for the "focus on my unit" action ... probably the svg obscuring the onclick

PR #858, opened 2023-11-02
fix onclick for focus-my-unit

chrisfarms:
* onclick for the name label was masking the clicks to the unit icon (click to focus on unit)
* resolves: #847


5p0rt5BEArD (approved):
LGTM and tested

Issue #736, opened 2023-10-19
We should deselect any selected buildings if I press the attack intent button


ldunnplaymint:
If the user walks up to an attack target, selects their unit and selects the target, then selects the attack intent button, they will notice that the attack button does not light up. The user has to re-select the enemy before the attack button will light up.

PR #860, opened 2023-11-03
deselect map element on intent switch

chrisfarms:
* resolves: #736


hypnoshock (approved):
Looks o.k but I haven't spun this up to test properly.

5p0rt5BEArD:
The title of the bug is fixed
The description of the bug is not.
Players still have to reselect buildings if they've selected them before pressing attack. Its a bit clearer they need to do that now but do we need to keep the bug alive with a different name? (if we do its not a high priority)

chrisfarms:
> The title of the bug is fixed The description of the bug is not. Players still have to reselect buildings if they've selected them before pressing attack. Its a bit clearer they need to do that now but do we need to keep the bug alive with a different name? (if we do its not a high priority)

nah it's just lazy triaging renaming as we go 😄

PR #861, opened 2023-11-03
allow selecting text in docs

chrisfarms:
fixes css preventing text selection in docs content ... which was annoying


hypnoshock (approved):
LGTM

Issue #864, opened 2023-11-03
ds-cli: apply not working for windows users


chrisfarms:
* screen clearing not playing nice in powershell
* apply -R not doing anything

PR #869, opened 2023-11-03
ds-cli: fix --recursive glob on windows

chrisfarms:
* glob paths must always be posix, even on windows (🤷)
* resolves: #864 

TODO:

* [x] get confirmation from a windows user that `ds apply -R -f .` works

this has already been published as v0.0.15


hypnoshock (approved):
TIL! Nicely fixed

PR #871, opened 2023-11-03
basic factoryspec validation on export

chrisfarms:
* run the manifest validation for the FactorySpec on click export
* resolves: #865

### DIFF
--- a/cli/src/utils/manifest.ts
+++ b/cli/src/utils/manifest.ts
@@ -26,6 +26,13 @@ export const Atom = z.number().gte(0).lt(4294967295);
 export const Slot = z.object({
     name: z.string(),
     quantity: z.number().gte(0).lte(100),
+}).superRefine(({ name, quantity }, ctx) => {
+    if (quantity && quantity > 0 && !name) {
+        ctx.addIssue({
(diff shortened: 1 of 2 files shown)


hypnoshock (approved):
👍

5p0rt5BEArD (approved):
minor comments
 - white space removed from exported basicFactory.js and not mentioned in commit message
 - its still possible to enter an item name of less than 3 chars and have an invalid yaml

tested and works
Fabricator doesn't allow empty item

image

ds apply with no item name error was 
`invalid manifest basic-factory.yaml: Item spec.name field invalid: String must contain at least 3 character(s)`
and is now
`invalid manifest basic-factory.yaml: BuildingKind spec.outputs.0 field invalid: Name required for item`

chrisfarms:
> * white space removed from exported basicFactory.js and not mentioned in commit message

sorry that'll be the auto linting on save 😇 

> its still possible to enter an item name of less than 3 chars and have an invalid yaml

yeah interesting, I'm using the actual manifest validation, so I guess we are not validating Slot spec name the same as Item spec name in the manifests ... noted, will try and catch it in future pr

Issue #808, opened 2023-10-31
Title for red goo tile doesn't match the visuals


PlaymintDavid:
Minor issue, but (to me) it would make more sense if the tooltip says "this tile has some red goo" on tiles where visually there is some red goo. I know that all (?) tiles have some amount of red goo so the tooltip is not inaccurate, just not intuitive.

<img width="869" alt="Screenshot 2023-10-31 at 09 42 22" src="https://github.com/playmint/ds/assets/129759334/cbc2ec85-c3ff-4ee0-a34c-70f4dd2c17d8">

chrisfarms:
yeah this is a bug ... only tiles that visually have the goo splodges should be displaying a "{RED/GREEN/BLUE} GOO TILE" just as you expect... the threshold for the tile description is probably out of date with some goo rate changes

PR #873, opened 2023-11-03
Comparing weights instead of rates to determine if a tile info displays goo name

hypnoshock:
# What

Comparing weights instead of rates to determine if a tile has some or lots of goo. By comparing weights we are applying the same logic that we use to display either a small or big goo texture.

image
image

Below shows tile with some blue goo but the tile info doesn't show the 'some goo' text
image

# Why

This solves the problem where a tile without a goo texture could say it had some goo on it. Even though this was technically true it didn't match the visuals

It might be that instead of comparing weights we make the textures display depending on rates. This would have the effect of showing more goo textures on the map. Not sure if that's what we wanted which is why I went with the weight option for now

resolves: #808

### DIFF
--- a/frontend/src/components/panels/tile-info-panel.tsx
+++ b/frontend/src/components/panels/tile-info-panel.tsx
@@ -10,7 +10,16 @@ import {
     WorldMobileUnitFragment,
 } from '@app/../../core/src';
 import { PluginContent } from '@app/components/organisms/tile-action';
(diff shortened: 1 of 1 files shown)


chrisfarms (approved):
yay fewer places to track rate changes

5p0rt5BEArD (approved):
tested

PR #886, opened 2023-11-06
Basic factory plus

### LINKED ISSUE #783: Example of a building that goes further than the creation QUEST
For example a Shop

5p0rt5BEArD:
### What

I've added functions to BasicFactory .js and .sol that gate the ability to craft. The calls to these functions have been commented out so the default exported versions from the fabricator are fully open crafting.

- log state (js)
- check if building is owned by unit owner (js + sol)
- check if building is authored by unit owner (js + sol)
- check if current player is on an allow list (js + sol)
- check if current player has an ID Card (sol only)

### Why

Shows a working example in code that could be used in a workshop scenario with the simplest code edit - a comment removal

resolves #783

### DIFF
--- a/contracts/src/example-plugins/BasicFactory/BasicFactory.js
+++ b/contracts/src/example-plugins/BasicFactory/BasicFactory.js
@@ -119,12 +119,12 @@ function unitIsFriendly(state, selectedBuilding) {
 }
 
 function unitIsBuildingOwner(mobileUnit, selectedBuilding) {
-    // console.log('unit owner id:',  mobileUnit?.owner?.id, `\nbuilding owner id`, selectedBuilding?.owner?.id);
+    // console.log('unit owner id:',  mobileUnit?.owner?.id, `building owner id`, selectedBuilding?.owner?.id);
     return mobileUnit?.owner?.id && mobileUnit?.owner?.id === selectedBuilding?.owner?.id;
 }
 
 function unitIsBuildingAuthor(mobileUnit, selectedBuilding) {
-    //console.log('unit owner id:',  mobileUnit?.owner?.id, `\nbuilding owner id`, selectedBuilding?.kind?.owner?.id);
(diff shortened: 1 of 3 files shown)


contracts/src/example-plugins/BasicFactory/BasicFactory.js
```diff
@@ -102,4 +103,37 @@ function inputsAreCorrect(state, building) {
...
+    return mobileUnit?.owner?.id && mobileUnit?.owner?.id === selectedBuilding?.kind?.owner?.id;
+}
+
+function unitOwnerConnectedToWallet(state, mobileUnit, walletAddress) {
+    //console.log('Checking player:',  state?.player, `vontrols unit`, mobileUnit, 'and has address:', walletAddress);
+    return mobileUnit?.owner?.id == state?.player?.id && 
+        state?.player?.addr == walletAddress;
+}
+
+// the source for this code is on githib where you can find other example buildings:
```
chrisfarms, line 137:
```suggestion
// the source for this code is on github where you can find other example buildings:
```

frontend/src/pages/building-fabricator.tsx
```diff
@@ -290,13 +371,13 @@ export default async function update(state) {
                         html: \`
                             <p>Fill the input slots to enable crafing</p>
                             \`,
-                       buttons: [
-                            {
-                                text: 'Craft',
-                                type: 'action',
-                                action: craft,
+                       buttons: [
```
chrisfarms, line 374: 👻  some weird whitespace at end of lines in this file

chrisfarms (approved):
LGTM 

I do think the intensionally commented out bits are clearer when you have added the "uncomment this to do X" notes next to them .... the bits that don't have this look a bit like dead code, but since this is mainly for walkthrough purposes I don't think it's a big deal either way

ldunnplaymint (approved):

PR #887, opened 2023-11-06
Fix outline scale at lower resolutions

JackPlaymint:
In order to improve performance on lower-end devices, Farms added in a resolution scaling system which helped to reduce the graphical load.
This worked well, but introduced a visual issue with the map element outlines, which would appear to be thicker than they're supposed to be.

resolving the issue turned out to be trickier than i expected, but the tldr is that I switched the resolution scaling to happen directly to the camera's target texture, rather than to the canvas web element.
This did have a side effect that meant that various Built-in Unity camera functions no longer worked as expected, so I had to implement custom versions of those functions to remedy that.

### DIFF
--- a/frontend/src/components/panels/nav-panel.tsx
+++ b/frontend/src/components/panels/nav-panel.tsx
@@ -97,7 +97,7 @@ export const NavPanel = ({
         if (!e) {
             return;
         }
-        const newHeight = parseInt(e.target.value, 10);
+        const newHeight = parseFloat(e.target.value);
         if (g.__globalUnityContext?.setCanvasHeight) {
             g.__globalUnityContext.setCanvasHeight(newHeight);
         }
@@ -195,13 +195,27 @@ export const NavPanel = ({
                         <fieldset>
                             <legend>Quality:</legend>
                             <select onChange={onChangeQuality} value={canvasHeight}>
-                                <option value="480">Low (480p)</option>
-                                <option value="720">Medium (720p)</option>
-                                <option value="1080">High (1080p)</option>
-                                <option value="-1">
+                                <option value={480 / Math.floor(window.innerHeight * window.devicePixelRatio)}>
+                                    Low (480p)
+                                </option>
+                                <option value={720 / Math.floor(window.innerHeight * window.devicePixelRatio)}>
+                                    Medium (720p)
+                                </option>
+                                <option value={1080 / Math.floor(window.innerHeight * window.devicePixelRatio)}>
+                                    High (1080p)
+                                </option>
+                                <option value={1440 / Math.floor(window.innerHeight * window.devicePixelRatio)}>
+                                    Ultra (1440p)
+                                </option>
+                                <option
+                                    value={
+                                        Math.min(window.innerHeight, window.innerHeight * window.devicePixelRatio) /
+                                        Math.floor(window.innerHeight * window.devicePixelRatio)
+                                    }
+                                >
                                     Auto ({Math.min(window.innerHeight, window.innerHeight * window.devicePixelRatio)}p)
                                 </option>
-                                <option value="-2">
+                                <option value="1">
                                     Native ({Math.floor(window.innerHeight * window.devicePixelRatio)}p)
                                 </option>
                             </select>
--- a/frontend/src/components/panels/nav-panel.tsx
+++ b/frontend/src/components/panels/nav-panel.tsx
@@ -92,7 +92,7 @@ export const NavPanel = ({
         window.location.reload();
     }, [clearSession, forgetProvider]);
 
-    const canvasHeight = g.__globalUnityContext?.getCanvasHeight ? g.__globalUnityContext.getCanvasHeight() : -1;
+    const canvasHeight = g.__globalUnityContext?.getCanvasHeight ? g.__globalUnityContext.getCanvasHeight() : 0.5;
     const onChangeQuality = useCallback((e) => {
         if (!e) {
             return;
--- a/frontend/src/components/panels/nav-panel.tsx
+++ b/frontend/src/components/panels/nav-panel.tsx
@@ -92,14 +92,14 @@ export const NavPanel = ({
         window.location.reload();
     }, [clearSession, forgetProvider]);
 
-    const canvasHeight = g.__globalUnityContext?.getCanvasHeight ? g.__globalUnityContext.getCanvasHeight() : 0.5;
+    const canvasHeight = g.__globalUnityContext?.getCanvasScale ? g.__globalUnityContext.getCanvasScale() : 0.5;
     const onChangeQuality = useCallback((e) => {
         if (!e) {
             return;
         }
         const newHeight = parseFloat(e.target.value);
-        if (g.__globalUnityContext?.setCanvasHeight) {
-            g.__globalUnityContext.setCanvasHeight(newHeight);
+        if (g.__globalUnityContext?.setCanvasScale) {
+            g.__globalUnityContext.setCanvasScale(newHeight);
         }
     }, []);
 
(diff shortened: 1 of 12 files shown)


frontend/src/components/panels/nav-panel.tsx
```diff
@@ -87,7 +87,7 @@ export const NavPanel = ({
         if (!e) {
             return;
         }
-        const newHeight = parseInt(e.target.value, 10);
+        const newHeight = parseFloat(e.target.value);
```
chrisfarms, line 90: Won't this cause it to attempt to update the canvas height to sub pixel sizes, and update more frequently that visible or am I misunderstanding the change?

frontend/src/hooks/use-unity-instance.tsx
```diff
@@ -65,28 +65,16 @@ const UnityInstance = () => {
...
+        canvas.width = mapContainerBounds.width * window.devicePixelRatio;
+        if (unity.sendMessage && g.__globalUnityContext.getCanvasHeight)
+            unity.sendMessage(
+                'ResolutionManager',
+                'SetResoltion',
+                JSON.stringify(g.__globalUnityContext.getCanvasHeight())
+            );
         console.info(`canvas size updated to ${canvas.width}x${canvas.height}`);
-    }, [canvas, canvasHeight]);
+    }, [canvas, unity]);
```
chrisfarms, line 77:
```suggestion
    }, [canvas, sendMessage]);
```

since the unity object will never change, this will never trigger an update ... switch to passing in the `sendMessage` func rather than the unity obj to make sure the sendMessage is the one you need

think of like this:

```
const o = {myFunc: () => 1};
const a = o;

o === a // true

o.myFunc = () => 2;

o === a // still true ... changing the myFunc did not change the fact that o is the same object.

**Review by chrisfarms — COMMENTED** [review]
Looks good ... I haven't spun it up to test yet 

... @ldunnplaymint be good if we could test on a lower power machine too to ensure we haven't regressed ... but maybe we just wait til deployed and remember to double check

File: frontend/src/hooks/use-unity-instance.tsx (hunk 2 above)
Comment (JackPlaymint) [reply frontend/src/hooks/use-unity-instance.tsx:77]: I see why I don't need unity, but I don't understand why I would to replace it with sendMessage, that wouldn't change either right?

**Comment by ldunnplaymint** [comment]
Getting errors when I build and run, rechecking but documenting them here

[message.txt](https://github.com/playmint/ds/files/13267874/message.txt)

File: frontend/src/hooks/use-unity-instance.tsx (hunk 2 above)
Comment (chrisfarms) [reply frontend/src/hooks/use-unity-instance.tsx:77]: it _might_ change ... since it comes from somewhere we do not control, so the hook requires

Comment (JackPlaymint) [reply frontend/src/hooks/use-unity-instance.tsx:77]: Ah gotcha! So adding the hook ensures that the sendMessage function will still work even if it gets changed somewhere down the road.

Comment (JackPlaymint) [reply frontend/src/hooks/use-unity-instance.tsx:77]: Linting complains about the useCallback having a missing dependency when I remove the unity hook:
`React Hook useCallback has a missing dependency: 'unity'. Either include it or remove the dependency array`

Should I just keep it in even as well as adding sendMessage, even though it will never change?

Comment (JackPlaymint) [reply frontend/src/hooks/use-unity-instance.tsx:77]: It's also complaining that sendMessage is unnecessary
`React Hook useCallback has an unnecessary dependency: 'sendMessage'. Either exclude it or remove the dependency array.`

Comment (chrisfarms) [reply frontend/src/hooks/use-unity-instance.tsx:77]: you need to not reference the `unity` variable inside the hook ... only reference the actual thing you need (in this case `sendMessage` which is already deconstructed at line ~53)

**Comment by ldunnplaymint** [comment]
I am still hitting this unreachable error. I did a git clean -fdx, a make clean, went to localhost and cleared my local storage, cleared my browsing data. I will ask Farms to give me a hand tomorrow and see what is happening. 

Also I rebased the ranch on Main

```Uncaught RuntimeError: unreachable
    at 0606635a:0x64760
    at 0606635a:0x57753
    at 0606635a:0x129ea80
    at 0606635a:0x12ba9ea
    at 0606635a:0xcfaf4c
    at Module.dynCall_iiiiiiiiii (a284c039-41af-4436-834b-579926b097b7:3:330368)
    at invoke_iiiiiiiiii (a284c039-41af-4436-834b-579926b097b7:3:396023)
    at 0606635a:0x1298f56
    at 0606635a:0x14378ba
    at 0606635a:0x10e987
    at 0606635a:0x3243cc
    at 0606635a:0xcfb460
    at invoke_iiii (a284c039-41af-4436-834b-579926b097b7:3:385610)
    at 0606635a:0x8fc6e
    at 0606635a:0x9e8cb9
    at 0606635a:0x648f8
    at 0606635a:0x17a0f9
    at 0606635a:0x1653a6
    at 0606635a:0xdcebdf
    at Module._SendMessageString (a284c039-41af-4436-834b-579926b097b7:3:311976)
    at Object.ccall (a284c039-41af-4436-834b-579926b097b7:3:11121)
    at Object.SendMessage (a284c039-41af-4436-834b-579926b097b7:3:471)
    at Object.SendMessage (ds-unity.loader.js:1:10001)
    at eval (use-unity-context.js:88:1)
    at eval (use-unity-instance.tsx:63:9)
    at eval (use-unity-instance.tsx:190:9)
    at commitHookEffectListMount (react-dom.development.js:23150:1)
    at commitPassiveMountOnFiber (react-dom.development.js:24926:1)
    at commitPassiveMountEffects_complete (react-dom.development.js:24891:1)
    at commitPassiveMountEffects_begin (react-dom.development.js:24878:1)
    at commitPassiveMountEffects (react-dom.development.js:24866:1)
    at flushPassiveEffectsImpl (react-dom.development.js:27039:1)
    at flushPassiveEffects (react-dom.development.js:26984:1)
    at performConcurrentWorkOnRoot (react-dom.development.js:25709:1)
    at workLoop (scheduler.development.js:266:1)
    at flushWork (scheduler.development.js:239:1)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js:533:1)```

File: frontend/src/hooks/use-unity-instance.tsx [hunk 3]
```diff
@@ -51,42 +51,17 @@ const UnityInstance = () => {
         productVersion: `blueprint`,
     });
     const { unityProvider, loadingProgression, addEventListener, removeEventListener, sendMessage } = unity;
-
-    const canvas = canvasRef.current ? (canvasRef.current as HTMLCanvasElement) : null;
-    const [canvasHeight, setCanvasHeight] = useLocalStorage<number>(`ds/canvasHeight`, -1);
+    const [canvasHeight, setCanvasHeight] = useLocalStorage<number>(`ds/canvasHeight`, 0.5);
```
chrisfarms, line 54: this values is no longer actually a canvasHeight ... I think it's some kind of quality percentage right?, so we should probably update all the references to this to reflect what it is to avoid confusion

frontend/src/hooks/use-unity-instance.tsx
```diff
@@ -51,42 +51,17 @@ const UnityInstance = () => {
...
-            // resolution, but never pack more pixels into the canvas than
-            // the actual screen pixels (ie if zoomed out or hiDPI)
-            // quirk: zooming IN will lower the res, which is weird, if you
-            // need to zoom in for some reason probably use "native"
-            canvas.height = Math.min(mapContainerBounds.height * window.devicePixelRatio, mapContainerBounds.height);
-            canvas.width = Math.min(mapContainerBounds.width * window.devicePixelRatio, mapContainerBounds.width);
-        }
-        console.info(`canvas size updated to ${canvas.width}x${canvas.height}`);
-    }, [canvas, canvasHeight]);
+        sendMessage('ComponentManager', 'SetResolution', canvasHeight != -1 ? canvasHeight : 0.5);
```
chrisfarms, line 63: sending the default 0.5 seems to crash unity? ... I have not built a debugmap yet to see the real error

chrisfarms:
closing this as it has gone stale

 there's nothing wrong with it, IIRC just needs a few tweaks but doesn't seem a priority at the moment

we can reopen if the fatty outlines is annoying people again later

PR #890, opened 2023-11-06
Added bag next to registration tower with IDs and Letters

hypnoshock:
# What

We can now spawn bags via yaml. I am spawning a bag that contains 100 x IDs and 100 x Registration letters

image

# Why

So if any pinches someone elses registration letter / ID during demonstration there are some spares they can take

### DIFF
--- a/cli/src/utils/applier.ts
+++ b/cli/src/utils/applier.ts
@@ -9,12 +9,31 @@ import {
     BuildingCategoryEnumVals,
     ContractSource,
     ManifestDocument,
-    Slot
+    Slot,
 } from '../utils/manifest';
 
 const null24bytes = '0x000000000000000000000000000000000000000000000000';
 
-export const encodeItemID = ({ name, stackable, goo }: {name: string, stackable: boolean, goo: {red: number, green: number, blue: number}}) => {
+export const encodeTileID = ({ q, r, s }: { q: number; r: number; s: number }) => {
+    return solidityPacked(
+        ['bytes4', 'uint96', 'int16', 'int16', 'int16', 'int16'],
+        [NodeSelectors.Tile, 0, 0, q, r, s]
+    );
+};
+
+export const encodeBagID = ({ q, r, s }: { q: number; r: number; s: number }) => {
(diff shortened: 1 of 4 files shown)


cli/src/utils/applier.ts
```diff
@@ -391,105 +406,148 @@ export const getOpsForManifests = async (
...
+            const quantities = [0, 0, 0, 0].map((_, idx) => (slots[idx] ? slots[idx].quantity : 0));
+            return { items, quantities };
+        };
+
+        const spec = doc.manifest.spec;
+        const [q, r, s] = spec.location;
+
+        const bagID = encodeBagID({ q, r, s });
+        const ownerAddress = solidityPacked(['uint160'], [0]); // public
+        console.log('ownerAddress: ', ownerAddress);
```
chrisfarms, line 529: stray debug log

chrisfarms (approved):
LGTM ... you've left a console.log in there which we prob don't want in the cli output

seems like there's some linting fighting going on we need to sort out for the cli ... not something for this PR tho

ldunnplaymint:
I am a bit worried about stacking them to 100. Most people will take the whole stack (Because it's left click to move the whole stack) so it feels like we will run into the same problem again anyway?

5p0rt5BEArD:
> I am a bit worried about stacking them to 100. Most people will take the whole stack (Because it's left click to move the whole stack) so it feels like we will run into the same problem again anyway?

Agree. 1 per spot I reckon. Maybe multiple bags.

hypnoshock:
> > I am a bit worried about stacking them to 100. Most people will take the whole stack (Because it's left click to move the whole stack) so it feels like we will run into the same problem again anyway?
> 
> Agree. 1 per spot I reckon. Maybe multiple bags.

Added more bags with 2 of each item

image

cli/src/utils/applier.ts
hypnoshock: Well spotted

ldunnplaymint (approved):

Issue #892, opened 2023-11-07
Clean up the empty QUEST progress bar


ldunnplaymint:
We should round off the quest progress bar when a quest has not had any progress 

![Image](https://github.com/playmint/ds/assets/99176798/2cadefec-af0b-45ec-971a-780c7dffe0fc)

PR #893, opened 2023-11-07
Quest shows a minimum of 5% complete

hypnoshock:
# What

We are now showing the progress bar at 5% if it's empty so that the zero state is a round orange circle

image

# Why

We want to show that the bar will fill up and so it matches the style of outlined bars it needs to be at least 5% so it's circular

resolves: #892


chrisfarms (approved):
🟠

ldunnplaymint (approved):
Orange you glad we did this change?

Issue #915, opened 2023-11-12
Deploy, not Apply


PlaymintDavid:
<img width="1040" alt="Screenshot 2023-11-12 at 15 34 32" src="https://github.com/playmint/ds/assets/129759334/936df27b-bc2c-443d-9562-eddd742c0330">

We should rename the Apply button Deploy, since a) deploying is cool, b) it says Deployed when you click it.

PR #916, opened 2023-11-13
label apply as deploy

chrisfarms:
rename "apply" button to "deploy"

resolves: #915


5p0rt5BEArD (approved):
Can't test right now but approving a one text string change

PR #927, opened 2023-12-08
Updated /ds README

Steedie:
- Updated for "Building from source (for development)" specifically for Windows. (Lewis to add macos version).
- Added collapsible sections to make it more organised considering the size of the updated guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,9 @@ See the [docs](./docs/README.md)
 
 ## How to build this project
 
-### Building with Docker
+<details>
+
+<summary>Building with Docker</summary>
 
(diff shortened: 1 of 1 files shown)


chrisfarms (commented):
Looks good nice one 👍 

One comment is that it does now look like we _only_ support Windows for development. Since that's what the section title says.

Could we either:

* Remove the "-Windows" suffix, so that it is a mixed guide both
* OR; Add an open/closey thing for "Building from source (macos/linux)" section too with the what-you-need+clone+make lines ... no expectation for it to be detailed, we can leave better macos instructions for another day

...basically just clarifying that we don't only support windows builds

whatcha think?

PR #928, opened 2023-12-08
Updated Makefile so it detects WSL

Steedie:
Updated the logic in the Makefile to detect whether the user is running on WSL, and adjusts the UNITY_EDITOR path accordingly.

Tested and is working on my Windows Machine (via WSL/Ubuntu). Perhaps should be tested to make sure everything still works like it should for macOS users.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,16 +1,20 @@
 
-
 # configure path to unity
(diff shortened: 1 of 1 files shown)


chrisfarms (approved):
looks good ... I don't have a windows box to test with WSL, but confirmed it still works on macos & linux

Issue #613, opened 2023-09-04
show name (if assigned) in OWNER and AUTHOR labels


chrisfarms:
in the little building/tile status label ... we should should the "name" instead of the address for the `AUTHOR` and `OWNER` labels

chrisfarms:
this will require we allow calling `NAME_OWNED_ENTITY` on Player kinds

5p0rt5BEArD:
Players don't have names (only units do)
So this probaby means adding a name entry field in the player connection dialogue window


![Image](https://github.com/playmint/ds/assets/90266137/2710d824-eba6-4470-8de7-b7ff6f6c735d)

PR #931, opened 2023-12-19
Bl 613 player names

Steedie:
# What

# Why

resolves: #613

### DIFF
--- a/frontend/src/components/panels/nav-panel.tsx
+++ b/frontend/src/components/panels/nav-panel.tsx
@@ -60,6 +60,7 @@ export const NavPanel = ({
     const { wallet } = useWallet();
     const player = usePlayer();
     const [showAccountDialog, setShowAccountDialog] = useState(false);
+    const [playerName, setPlayerName] = useState('');
 
     const hasConnection = player || wallet;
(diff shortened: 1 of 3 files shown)


frontend/src/components/panels/nav-panel.tsx
```diff
@@ -93,6 +102,26 @@ export const NavPanel = ({
...
+            console.warn("naming failed: no player");
+            return;
+        }
+        if (!playerName || playerName.length < 3) {
+            console.warn("naming failed: no name or length less than 3 characters");
+            return;
+        }
+        if (playerName.length > 20) {
+            console.warn("naming failed: max 20 characters");
+            alert('rejected: max 20 characters');
```
hypnoshock, line 116: Maybe put the error in a state object and render it on the form instead of using an alert

hypnoshock (approved):
Looks good so I won't hold up the merge but maybe render the error out to the form instead of displaying an alert. It's a minor thing though

frontend/src/components/panels/nav-panel.tsx
Steedie: Thanks for pointing this out, will get that added later today.

PR #935, opened 2024-01-09
Fix forge fmt

5p0rt5BEArD:
Matching the github action foundry version with the Dockerfile so that 'forge fmt' does not get out of sync

### DIFF
--- a/.github/workflows/check-contracts-test.yaml
+++ b/.github/workflows/check-contracts-test.yaml
@@ -23,7 +23,7 @@ jobs:
       - name: setup
         uses: foundry-rs/foundry-toolchain@v1
         with:
-          version: nightly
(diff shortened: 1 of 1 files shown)


5p0rt5BEArD:
OK lets not do this 
I've updated foundry locally to match the 'forge fmt' used but github actions

chrisfarms (commented):
when we pinned this previously it lead to very flakey checks as it wasn't always able to fetch the pinned version

I agree this _should_ be what we want ... but we moved away to avoid the flakes, so not really sure if this is better than the occasional `foundryup` to stay in sync

5p0rt5BEArD:
Closing this PR as something we do not want to do as it causes flakiness

PR #937, opened 2024-01-12
Added a duck icon to the fabricator set

5p0rt5BEArD:
### What
Additional icon added to the list that fabricator webpage allows you to pick from

image

### Why
Its a relevant icon given the Ducks v Burgers trend


chrisfarms (approved):
🦆

PR #940, opened 2024-01-12
tidy errors in create-a-new-item.md (OPEN QUESTION)

okwme:
This is the first time I've seen the word "atom" used anywhere and it's not really introduced. Is it interchangeable with goo?

### DIFF
--- a/docs/how-to/create-a-new-item.md
+++ b/docs/how-to/create-a-new-item.md
@@ -13,11 +13,11 @@ All items in Downstream contain goo. There are 3 types of goo:
 -   Blue Goo
 -   Red Goo
 
-When a new item is registered you state its Goo composition. However, these numbers can only by at most half the value of the goo in its input items.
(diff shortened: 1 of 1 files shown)


5p0rt5BEArD:
Good spot. An earlier version of the game referred to atoms rather than goo.

Suggested edits would be

line 20
`The easiest way to see how much goo an item has is to hover over it in the inventory. The base resources found in the game contain:`

lines 36 and 37
```
-   Stackable items can be stacked, so that up to 100 fit into 1 inventory slot. However, although they have goo, the goo amounts are not taken into account when calculating combat stats
-   Equipable items can not be stacked. Only 1 will fit into 1 slot. These items DO have their goo amounts taken into account during combat. (see [Combat stats](/docs/how-to/play-the-game#combat-stats))
```

okwme:
sounds good @5p0rt5BEArD !
i realized my first very small edit was also missing so put it back in as well as your suggestions

5p0rt5BEArD (approved):

ldunnplaymint (approved):

PR #941, opened 2024-01-12
Building construction block recorded as edge

hypnoshock:
# What

Building's construction block number now recorded on a `BlockNum` edge instead of using data storage

# Why

Data storage for a node should be reserved for user data and not include any default data

# Note

We were recording block times for exaction and using a `BlockNum` node connected using a generic `Has` edge. I have added an explicit `BlockNum` edge to avoid problems where the same edge type could be used for something else. As there is now the possibility of type block numbers being recorded for a building (construction and extraction) I have added an enum for the keys where 0 is used for construction and 1 is used for extraction. 

I have updated graphQL to use the new keys and tested extraction on the map which appeared to work as it did before


5p0rt5BEArD (approved):
Looks good to me and I've had a quick test to see the state storage and extractor buildings still working

ldunnplaymint (approved):

PR #955, opened 2024-01-16
DB HQ finds your green goo

5p0rt5BEArD:
DB HQ looks for green goo in all Unit slots and only enables the join button if you have it.
Joining then takes it from the first slot it found it in.


ldunnplaymint (approved):

ldunnplaymint:
Looks good. I messed around with lots of weird inventory set ups and it seems solid

PR #963, opened 2024-01-17
Only counts local ducks and burgers for local units

JackPlaymint:
This PR makes the DvBHQ only count duck and burger buildings that are inside its 5 tile radius.


5p0rt5BEArD (approved):
Approved. That's what I was going to do :)

PR #964, opened 2024-01-17
Prizes are now transferred to the first empty or valid slot

hypnoshock:
# What

Prize destination item slot is now automatically found instead of hard coded to bag 0 slot 0. Prizes are transferred to the first empty or same type slot with sufficient capacity.

image
One valid empty slot

image
Empty slot is filled with reward on first claim

image
Slot of same type found on second claim so balance increased from 2 to 4


5p0rt5BEArD (approved):
Nice.
You make shake your fist at stack too deep but its probably a good readability refactor :)

PR #965, opened 2024-01-18
1kx/attack defense buildings

joergkiesewetter:
trying to add 2 new features:

- a hook while building, to add new mechanics like conditional building and consumable items
- trying to add a new building category called 'military'. It should behave like a factory, but without input and output ressources

### DIFF
--- a/cli/src/commands/get.ts
+++ b/cli/src/commands/get.ts
@@ -129,6 +129,7 @@ const nodeToManifest = (node): z.infer<typeof Manifest> => {
         };
         const model = node.model?.value || '';
         // TODO: fixup once category and extractor is a real thing, this is currently a guess
+        // TODO [military] check for military category
         const category = outputs.length > 0 ? 'factory' : implementation ? 'custom' : 'blocker';
         const spec = (() => {
             if (category === 'factory') {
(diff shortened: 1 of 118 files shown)


5p0rt5BEArD:
> trying to add a new building category called 'military'. It should behave like a factory, but without input and output ressources

Not looked through code yet but can you explain the intension a bit more here?

Buildings already take part in combat if a Unit attacks it or stands on its tile and attacks an adjacent building.

is the idea here that the building can start the combat itself?

joergkiesewetter:
the idea is to have a building which acts as a factory on map, means you can build it, and walk on it when you want to attack. But we don't need a building for these buildings, they should only contribute to your combat stats.

I think for now this is more likea QoL think, but in later iterations it could make sense to have special abilities on military buildings we don't want to have on factories.

But I'm also happy to split this PR up into two distinct, so that the first functionality could get deployed.

joergkiesewetter:
sounds good for me, I created a new PR with just the construction hook https://github.com/playmint/ds/pull/967

I think for the military buildings, we can postphone this, since there are workarounds available

chrisfarms:
👍 ... closing in favour of #967

PR #967, opened 2024-01-19
Construction Hook

joergkiesewetter:
adding a construction hook to the building kind interface

### DIFF
--- a/contracts/src/rules/BuildingRule.sol
+++ b/contracts/src/rules/BuildingRule.sol
@@ -231,6 +231,16 @@ contract BuildingRule is Rule {
         if (TileUtils.distance(mobileUnitTile, targetTile) > 1 || !TileUtils.isDirect(mobileUnitTile, targetTile)) {
             revert("BuildingMustBeAdjacentToMobileUnit");
         }
+
+        // calls the preflight hook in the building implementation to add the possibility to make conditional builds 
+        // or consume inventory while building
+        BuildingKind buildingImplementation = BuildingKind(state.getImplementation(buildingKind));
+        // if no implementation set, then this is a no-op
+        if (address(buildingImplementation) == address(0)) {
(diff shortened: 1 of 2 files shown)


contracts/src/rules/BuildingRule.sol
```diff
@@ -231,6 +231,16 @@ contract BuildingRule is Rule {
...
         }
+
+        // calls the preflight hook in the building implementation to add the possibility to make conditional builds 
+        // or consume inventory while building
+        BuildingKind buildingImplementation = BuildingKind(state.getImplementation(buildingKind));
+        // if no implementation set, then this is a no-op
+        if (address(buildingImplementation) == address(0)) {
+            return;
+        }
+        buildingImplementation.construct(game, buildingKind, mobileUnit, abi.encode(coords));
```
chrisfarms, line 242: Some buildings don't have contracts ... but you still want them to get built ... so bailing earlier here will prevent those kinds of buildings ever getting contracted ... I think you want something like:

```suggestion
        if (address(buildingImplementation) != address(0)) {
            buildingImplementation.construct(game, buildingKind, mobileUnit, abi.encode(coords));
        }
```

5p0rt5BEArD (approved):

chrisfarms:
.... I'm just trying to work out a way this can get in without requiring breaking (or requiring updating of) existing all existing BuildingKinds 😅

5p0rt5BEArD:
> .... I'm just trying to work out a way this can get in without requiring breaking (or requiring updating of) all existing BuildingKinds 😅

Ah ok - I was assuming all was ok as this was already iterated on in the original PR

My approval was hasty

joergkiesewetter:
sorry, didn't thought about this. would it make sense to make BuildingKind an abstract class?

PR #978, opened 2024-01-19
Fixing up BuildingKinds to support on-construct hook

chrisfarms:
had to make BuildingKind an abstract contract to allow other BuildingKind implementation to only optionally implement the `construct` callback

related: https://github.com/playmint/ds/pull/967


5p0rt5BEArD (approved):
Looks correct and have tested a couple of buildings on local build (and ran contracts tests)

PR #979, opened 2024-01-19
Duck Burger HQ - No longer relies on teamDuck/BurgerUnits array

Steedie:
### What

1. The code has been optimized and refined.
2. The Duck Burger HQ building no longer relies on the teamDuckUnits and teamBurgerUnits array for it to work, it now correctly uses the building instance's data.

### Why

1. Otherwise, I was getting this error stopping me from deploying the building: 
> "Warning: Contract code size is 25281 bytes and exceeds 24576 bytes (a limit introduced in Spurious Dragon). This contract may not be deployable on Mainnet. Consider enabling the optimizer (with a low "runs" value!), turning off revert strings, or using libraries"
2. So that multiple Duck Burger HQs can be used at the same time.


Steedie:
Just re-added the comments and made a small fix to setting the endBlock - ready for review.

PR #994, opened 2024-01-24
Performance Testing

Steedie:
### What & Why
We want a process for testing the performance of Downstream. The objective of this PR is to develop a way of being able to test performance to discover potential issues that may occur at larger scales of gameplay. For example: 

- When there are a greater number of "Always Active" plugins running in the session.
- When there is a greater number of units moving around at the same time.

As a result of this work, I want the tester to be able to spin up a session with a configurable amount of "Always Active" plugins that are deployed upon start up, and for the map to be programmatically generated, and place the deployed plugins in the map automatically.

### How
What I've done to implement the above objective is create a new Makefile command `Make performance_test` that prompts the tester to enter an amount of Duck Burger games to set up, and passes the input through to .performance_test_startup.js:
```Make
performance_test: all
	@echo "Enter number of Duck Burger HQ buildings to create: "; \
	read number; \
	$(NODE) .performance_test_startup.js $$number
```

.performance_test_startup.js then programmatically creates new Duck Burger related yaml files, copies the sol and js files, and moves them over to a plugin-test folder in `contracts/src/`. It then builds the map, and places the newly created plugins on it, and then runs `ds apply` to deploy everything.

This is the result of running `Make performance_test` with an input of `3`:
image

### What's Next

- [x] Clean up the current implementation of the code where possible
- [x] Begin testing with chaos units to discover potential problems
- [ ] Identify what causes the problems _(in progress)_


contracts/src/example-plugins/DuckBurger/DuckBurgerHQ.js
```diff
@@ -121,11 +121,12 @@ export default async function update(state) {
     // so we need to find all HQs on the map and update them each in turn
     //
     // for now we just update the first we find
-    const dvbBuildingName = "Duck Burger HQ";
+    const dvbBuildingName = "Duck Burger HQ".toLowerCase();
     const selectedBuilding = state.world?.buildings.find(
-        (b) => b.kind?.name?.value == dvbBuildingName,
+        (b) => b.kind?.name?.value.toLowerCase().includes(dvbBuildingName),
```
Steedie, line 126: Added .includes here because the new BuildingKind(s) add "_X" to the end of the names.

Steedie: The change made in my previous comment made it so only the first Duck Burger game would work. This has now been resolved by changing `dvbBuildingName` from `.teststartup.js`, therefore allowing multiple games to be played at the same time:
```js
function modifyJsContent(content, index) {
    return content.replace(/(Burger Display Building|Duck Display Building|Countdown Building|Duck Burger HQ)/g, `$1_${index}`);
}
```

chrisfarms (commented):
Chatted about this PR IRL ...

generally LGTM ... does what we want and works as advertised.

but the duplication of code between devstartup and teststartup scripts and the makes it very likely this code will get stale so we agreed we'd like to see:

* the manifest generation stuff broken out into a script (which could live in `./scripts/`)
* some kind of way to tell the main devstartup script that you want to generate and load this performance-testing map ... something like running `MAP=performance-test make dev` to select it for local deployments

Steedie:
> Chatted about this PR IRL ...
> 
> generally LGTM ... does what we want and works as advertised.
> 
> but the duplication of code between devstartup and teststartup scripts makes it very likely this code will get stale so we agreed we'd like to see:
> 
> * the manifest generation stuff broken out into a script (which could live in `./scripts/`)
> * some kind of way to tell the main devstartup script that you want to generate and load this performance-testing map ... something like running `MAP=performance-test make dev` to select it for local deployments

Thanks for the direction here, this has now been addressed in my latest commit. I've removed the unnecessary "performance-test"-related command/startup file, and have added optional env vars: `MAP`, and `ARENAS` to the `make dev` command:
```Makefile
dev: all
ifeq ($(MAP),)
	$(NODE) .devstartup.js
else
	$(NODE) .devstartup.js $(MAP) $(ARENAS)
endif
``` 
and moved the logic for generating the performance test to the scripts folder.

I've tested this new implementation (`make dev`, `make dev MAP=performance-test ARENAS=3`, `make dev MAP=tiny` etc...), and that's all working correctly. However, I've noticed a slight problem: I'm getting this error:
image
when I run `make dev` / `make dev MAP=tiny` after deleting the `test-plugins` folder (created after running `make dev MAP=performance-test`), but works again after I run a `make clean`, or regenerate the test files.

I'm wondering if you know why the old path is still trying to be accessed by something. I'm assuming it's running some kind of check somewhere else but breaks since the path doesn't exist anymore?

chrisfarms:
> I'm wondering if you know why the old path is still trying to be accessed by something. I'm assuming it's running some kind of check somewhere else but breaks since the path doesn't exist anymore?

I think this is `forge` caching things, and then getting upset when you delete things .... I believe `forge build`is not going to get rerun if all you did was delete your test-plugins

... you could check this is the case by running `cd contracts && forge cache clean && forge build` after you delete you generated test-plugins folder, and see if that fixes it

If that is the case, then you might need to force `make dev` to run the build step more frequently, something like making it a dependency of the `dev:` target like:

```
dev: all contracts
	$(NODE) .devstartup.js
```

... not ideall as it might slow down the startup for all the cases where it didn't really need to do a rebuild, but we might just have to live with it

Steedie:
> > I'm wondering if you know why the old path is still trying to be accessed by something. I'm assuming it's running some kind of check somewhere else but breaks since the path doesn't exist anymore?
> 
> I think this is `forge` caching things, and then getting upset when you delete things .... I believe `forge build`is not going to get rerun if all you did was delete your test-plugins
> 
> ... you could check this is the case by running `cd contracts && forge cache clean && forge build` after you delete you generated test-plugins folder, and see if that fixes it
> 
> If that is the case, then you might need to force `make dev` to run the build step more frequently, something like making it a dependency of the `dev:` target like:
> 
> ```
> dev: all contracts
> 	$(NODE) .devstartup.js
> ```
> 
> ... not ideall as it might slow down the startup for all the cases where it didn't really need to do a rebuild, but we might just have to live with it

Ahh got it now. I tried `cd contracts && forge cache clean && forge build` and it didn't work, but then I ran **only** `rm -rf contracts/cache` from `make clean` and it fixed the issue! So I suppose in this case we can run that every `make dev` instead of `contracts`?

chrisfarms:
> Ahh got it now. I tried `cd contracts && forge cache clean && forge build` and it didn't work, but then I ran **only** `rm -rf contracts/cache` from `make clean` and it fixed the issue! So I suppose in this case we can run that every `make dev` instead of `contracts`?

give it a go! ... something feels a bit wrong about pulling the cache from forge without then also calling build to rebuild it, but forge is probably smart enough to rebuild stuff anyway so maybe I'm overthinking it

Steedie:
> > Ahh got it now. I tried `cd contracts && forge cache clean && forge build` and it didn't work, but then I ran **only** `rm -rf contracts/cache` from `make clean` and it fixed the issue! So I suppose in this case we can run that every `make dev` instead of `contracts`?
> 
> give it a go! ... something feels a bit wrong about pulling the cache from forge without then also calling build to rebuild it, but forge is probably smart enough to rebuild stuff anyway so maybe I'm overthinking it

Working now like this:
```Makefile
dev:
	rm -rf contracts/cache
	$(MAKE) all
ifeq ($(MAP),)
	$(NODE) .devstartup.js
else
	$(NODE) .devstartup.js $(MAP) $(ARENAS)
endif
```
It's taking me an extra ~30 seconds to rebuild the cache every time which is a shame. Perhaps when the test-plugins folder is deleted (when setting `ARENAS` to 0) it can target the cache and rip out the reference to the deleted paths. I'll have a look if I can do this.

chrisfarms:
Looking at this afresh .... I don't think we should have the `rm` line in the `dev` target.

I actually think it's fine that it breaks if you delete the files .... there are many cases where the same thing is true, requiring you to do a `make clean` if you have messed with stuff, and so I don't think we should treat this any differently.

We should however ensure that:

* the `test-plugins` directory is deleted as part of `make clean`
* the `test-plugins` directory is in the`.gitignore` file so they don't show up as modified files

Steedie:
> Looking at this afresh .... I don't think we should have the `rm` line in the `dev` target.
> 
> I actually think it's fine that it breaks if you delete the files .... there are many cases where the same thing is true, requiring you to do a `make clean` if you have messed with stuff, and so I don't think we should treat this any differently.
> 
> We should however ensure that:
> 
> * the `test-plugins` directory is deleted as part of `make clean`
> * the `test-plugins` directory is in the`.gitignore` file so they don't show up as modified files

Ah ok that sounds like the most reasonable solution, I'll work on these changes.

chrisfarms (approved):

PR #1000, opened 2024-01-26
Max out the goo values and remove the map splats

JackPlaymint:
This PR disables the random perlin distribution of goo on the map and instead maximises the amount of goo on each tile evenly.
It also removes the creation of TileGoo components so the map isn't covered in goo splats.


chrisfarms (approved):
here's a cake and approval for being the 1000th PR customer 🍰 ✅

Issue #1013, opened 2024-02-02
ds-cli: You cannot apply Building manifest unless you also pass the BuildingKind manifest


ldunnplaymint:
when you deploy a `BuildingKind` manifest that references an Item name ... the cli applier attempts to find the referenced Item either in the collection of manifests currently being applied OR by looking at the set of already deployed Items

but when you deploy a `Building` manifest that references a BuildingKind... the cli ONLY looks at the collection of manifests currently being applied.

this means that **it currently fails to apply a `Building` manifest unless you are also deploying the `BuildingKind` manifest at the same time.**

it should probably work like the first example, and not fail.

PR #1019, opened 2024-02-05
Buildings can now be deployed without supplying the kind manifest

hypnoshock:
# What

Fixes the problem where buildings couldn't be deployed unless the `buildingKind` for that that building was deployed at the same time (and existed in the pendingManifest list).

# Fix

This was due to the passed in `existingBuildingKinds` in the `getBuildingKindIDByName` function being treated as specs when they were the graphQL fragments

# Test

I tested this by deploying some small rocks:

```
---
kind: Building
spec:
  name: Small Rocks
  location: [1, 0, -1]

```

fixes: #1013


chrisfarms (approved):

5p0rt5BEArD (approved):
Tested and is fixed for me locally

PR #1020, opened 2024-02-05
Hide units if they haven't moved for 30 minutes

### LINKED ISSUE #972: Remove Unit corpses
Units stay on the map forever which is not good for repeating experiences on the same map with different players (or just new burners)

Not sure of proper solution but it could be a frontend cheat for now

- Units that have not been interacted with for X blocks are not shown. Filtered out by graphql into a separate list

bonus

- PLayer account dialogue shows any dead units and you can revive them

bonus (or optimisation)

- there is an action to delete all "dead" units from onchain state

hypnoshock:
# What

If a unit hasn't moved for ~30 minutes (900 blocks) then hide them on the map unless they are the player's unit (to allow for selection)

# Why

A map can become littered with 'dead' units if multiple game sessions have been played within one deployment

# Details

The time can be adjusted using the `UNIT_DISPLAY_TIMEOUT_SECS` constant in `src/components/map/MobileUnit.tsx`

Fixes #972

### DIFF
--- a/frontend/src/components/map/ExtractorBuilding.tsx
+++ b/frontend/src/components/map/ExtractorBuilding.tsx
@@ -1,3 +1,4 @@
+import { BLOCK_TIME_SECS } from '@app/fixtures/block-time-secs';
 import { getGooPerSec } from '@app/helpers/tile';
 import { useBlock } from '@app/hooks/use-game-state';
 import { UnityComponentProps, useUnityComponentManager } from '@app/hooks/use-unity-component-manager';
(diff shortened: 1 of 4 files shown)


chrisfarms (approved):
Looks like it does what we want.

It it a problem that this invalidates the entire `units` memo on every block? 🤔 

Could maybe keep track of the "oldest" unit, and only invalidate the memo based on that or something .... but maybe that's just overcomplicating things until there is proof that the perf gain is worth it.... so it's probably fine

ldunnplaymint (approved):

PR #1022, opened 2024-02-05
cli 0.0.17

5p0rt5BEArD:
### What
https://www.npmjs.com/package/@playmint/ds-cli v 0.0.17 was published and these are the package file changes 

### WHy
v0.0.16 was not working with latest build, at least because the buildKind interface has changed

### How
Ran `make publish` and committed the changes


chrisfarms (approved):
good catch ... and reminds me that we will need to publish again after #1019 is in \cc @hypnoshock

PR #1025, opened 2024-02-06
Performance - Gracefully Handle OOM

### LINKED ISSUE #1001: Gracefully handle OOM in sandbox
We are currently having OOM issues causing plugins to fall over in sandbox

Steedie:
## What
While testing, we have discovered that when the sandbox has run out of memory, all plugins fall over causing indefinite errors, and plugins to fail.
We want a way of catching the out-of-memory errors as they appear, and immediately addressing them before they lead to more problems (plugins/sandbox to crash).

So far, there are certain checks similar to this:
```js
catch (err) {
                if (String(err).includes('memory')) {
                    try {
                        await disposeRuntime();
...
```
that call the disposeRuntime() function:
```js
export async function disposeRuntime() {
    // renamed branch
    if (!runtime) return;
    console.log('%c DISPOSING CONTEXTS & RUNTIME...', 'background: #222; color: #bada55');
    const disposeContextPromises = contexts.map((ctx) => {
        return new Promise((resolve, reject) => {
            try {
                ctx.dispose();
                resolve(ctx);
            } catch (err) {
                reject(err);
            }
        });
    });

    try {
        await Promise.all(disposeContextPromises);
        console.log('%c contexts disposed successfully...', 'background: #222; color: #bada55');
    } catch (err) {
        console.error('Error disposing contexts: ', err);
    }
    //contexts = [];

    try {
        clearTimeout(pollPendingJobsTimeout);
        console.log('%c cleared pending jobs timeout (remove this later?)...', 'background: #222; color: #bada55');
    } catch (err) {
        console.error('Error clearing pollPendingJobs timeout: ', err);
    }
    try {
        // console.log('runtime has pending stuff? ', runtime.hasPendingJob());
        console.log('runtime: ', runtime);
        runtime.dispose(); // Error disposing runtime:  RuntimeError: Aborted(Assertion failed: list_empty(&rt->gc_obj_list), at: quickjs/quickjs.c,1983,JS_FreeRuntime).
        console.log('%c runtime disposed successfully...', 'background: #222; color: #bada55');
    } catch (err) {
        console.error('Error disposing runtime: ', err);
    }
}
```

The current dilemma is that executing `runtime.dispose()` is triggering this error:
_"Error disposing runtime:  RuntimeError: Aborted(Assertion failed: list_empty(&rt->gc_obj_list), at: quickjs/quickjs.c,1983,JS_FreeRuntime)."_

I am currently working to resolve this error so I can make a way to reboot the runtime/contexts

_Edit 09/02/24_:
Worked with Farms on this issue, and we're now restarting the sandbox whenever OOM is detected!
We are still disposing and removing contexts (plugins) if they trigger an error, but the `disposeRuntime` function is no longer needed.

resolves #1001

### DIFF
--- a/cli/src/commands/test.ts
+++ b/cli/src/commands/test.ts
@@ -67,6 +67,12 @@ const chaosUnit = {
             [q,r,s] = validNeighbours[Math.floor(Math.random() * validNeighbours.length)] || [0,0,0];
 
             await player.dispatch({ name: 'MOVE_MOBILE_UNIT', args: [unitKey, q, r, s] }).then(res => res.wait());
+            // if (Math.random() > 0.75){
+            //     const building = Math.random() > 0.5 ? "0xbe92755c0000000000000000546391e80000000000000003" : "0xbe92755c0000000000000000444749c70000000000000003";
+            //     const args = { name: 'CONSTRUCT_BUILDING_MOBILE_UNIT', args: [unitId, building, q, r, s] };
+            //     await player.dispatchAndWait(args);
+            //     // failed to call: execution reverted: no construction bag found
+            // }
             await sleep(Math.floor(Math.random() * 2000)); // jitter
         }
     },
--- a/cli/src/commands/test.ts
+++ b/cli/src/commands/test.ts
@@ -67,12 +67,6 @@ const chaosUnit = {
             [q,r,s] = validNeighbours[Math.floor(Math.random() * validNeighbours.length)] || [0,0,0];
 
             await player.dispatch({ name: 'MOVE_MOBILE_UNIT', args: [unitKey, q, r, s] }).then(res => res.wait());
-            // if (Math.random() > 0.75){
-            //     const building = Math.random() > 0.5 ? "0xbe92755c0000000000000000546391e80000000000000003" : "0xbe92755c0000000000000000444749c70000000000000003";
-            //     const args = { name: 'CONSTRUCT_BUILDING_MOBILE_UNIT', args: [unitId, building, q, r, s] };
-            //     await player.dispatchAndWait(args);
-            //     // failed to call: execution reverted: no construction bag found
-            // }
             await sleep(Math.floor(Math.random() * 2000)); // jitter
         }
     },
--- a/core/src/plugins.ts
+++ b/core/src/plugins.ts
@@ -37,6 +37,7 @@ import { getBagsAtEquipee, getBuildingAtTile } from './utils';
 import { Logger } from './logger';
 
 const active = new Map<string, ActivePlugin>();
+//
 
 /**
  * makeAvailablePlugins polls for the list of deployed plugins every now and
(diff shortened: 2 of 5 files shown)


chrisfarms:
If you are still struggling with getting quickjs to be disposed cleanly ... maybe we could destroy the entire WebWorker process instead!

This hook is what provisions the web worker with the sandbox in it: https://github.com/playmint/ds/blob/main/frontend/src/hooks/use-game-state.tsx#L83-L95

If we were to expose some way of invalidating that useEffect to cause it to re-run, then it could do what you are trying to do.

we could extend the DSContextValue game state context to return a `resetSandbox` func....

```
 const [numSandboxRestarts, setSandboxRestarts] = useState(0);
 const restartSandbox = useCallback(() => setSandboxRestarts(prev => prev + 1), []);

 useEffect(() => {
        workerRef.current = new Worker(new URL('../workers/sandbox.ts', import.meta.url));
        const workerSandbox: Comlink.Remote<Sandbox> = Comlink.wrap(workerRef.current);
        workerSandbox
            .init()
            .then(() => {
                setSandbox({ sandbox: workerSandbox });
            })
            .catch(() => console.error(`sandbox init fail`));
        return () => {
            workerRef.current?.terminate();
        };
    }, [numSandboxRestarts]); // <---- invalidate the useEffect when the num incremened

    ....

    const value = useMemo(() => ({...sources, restartSandbox}), [sources, restartSandbox]);
    return <DSContext.Provider value={value}>{children}</DSContext.Provider>;
```


then from somewhere else we can call the fun:

```
const { restartSandbox } = useGameState();
restartSandbox(); // blow up the worker
```


... but then you would have to problem of how to know when you need to call restartSandbox() 🤔  .... hmmmmm maybe that's harder

Steedie:
> If you are still struggling with getting quickjs to be disposed cleanly ... maybe we could destroy the entire WebWorker process instead!
> 
> This hook is what provisions the web worker with the sandbox in it: https://github.com/playmint/ds/blob/main/frontend/src/hooks/use-game-state.tsx#L83-L95
> 
> If we were to expose some way of invalidating that useEffect to cause it to re-run, then it could do what you are trying to do.
> 
> we could extend the DSContextValue game state context to return a `resetSandbox` func....
> 
> ```
>  const [numSandboxRestarts, setSandboxRestarts] = useState(0);
>  const restartSandbox = useCallback(() => setSandboxRestarts(prev => prev + 1), []);
> 
>  useEffect(() => {
>         workerRef.current = new Worker(new URL('../workers/sandbox.ts', import.meta.url));
>         const workerSandbox: Comlink.Remote<Sandbox> = Comlink.wrap(workerRef.current);
>         workerSandbox
>             .init()
>             .then(() => {
>                 setSandbox({ sandbox: workerSandbox });
>             })
>             .catch(() => console.error(`sandbox init fail`));
>         return () => {
>             workerRef.current?.terminate();
>         };
>     }, [numSandboxRestarts]); // <---- invalidate the useEffect when the num incremened
> 
>     ....
> 
>     const value = useMemo(() => ({...sources, restartSandbox}), [sources, restartSandbox]);
>     return <DSContext.Provider value={value}>{children}</DSContext.Provider>;
> ```
> 
> then from somewhere else we can call the fun:
> 
> ```
> const { restartSandbox } = useGameState();
> restartSandbox(); // blow up the worker
> ```
> 
> ... but then you would have to problem of how to know when you need to call restartSandbox() 🤔 .... hmmmmm maybe that's harder

Ok thanks for this suggestion! There's a few more things I'd like to try but I'll give this a go if I can't get anything working.

Steedie:
Problem no longer a problem!
```js
try {
        await init();
```
Instead of trying to dispose the runtime, I'm now just running init() again to overwrite it. I've tested this by calling runtime.dumpMemoryUsage() before and after doing this and it is indeed reset to its initial state.

Now I'm looking into re-loading the plugins into the newly remade runtime, which has already led me to `use-game-state.tsx`...

Steedie:
Worked with Farms today on the remaining issues -  problems are now solved. When we detect an *out of memory* error, the sandbox is restarted, rebuilding the plugins.

PR ready for review

chrisfarms:
rebased on main

chrisfarms (approved):
played around with it this morning, looks like it's behaving as well as can be expected.

shame the code is so all over the place, but that's not a reflection on this PR, more a reflection on how the sandbox/plugins stuff is split accross core + frontend ... we should consider moving all the plugin related stuff out of core and into frontend in the future.

✅

PR #1026, opened 2024-02-06
Empty map, manual manifest apply, reorganised map folder structure

JackPlaymint:
This PR is a bit of a doozy, it moves a lot of files and folders around and will definitely need checking over.

To start, I created a new map folder called 'default' which contains a map.yaml file which creates a single tile only.
I then bypassed the restrictions in CheatsRule.sol by including a boolean set to true and checking for it before doing the usual authorisation checks.

With those changes, the almost empty default map will load by default, and then anyone can manipulate the world using ds apply without restriction.

I then rearranged the maps folder structure, removing the 'fixtures' folder entirely.
Instead what was in the fixtures folder now lives in maps/quest-map and can be applied to the world with ds-apply.

I also added a 'src/common' folder which contains some common elements that might be shared between maps. for now, this contains blockers, extractors and the allowlist.json.
These common items can then be accessed using symlinks as demonstrated in the quest-map's Decorations and extractors folders.


ldunnplaymint (changes requested):
I didn't meant to approve this before, my bad! When I just run a make dev, I get a one hex map!

chrisfarms:
> I didn't meant to approve this before, my bad! When I just run a make dev, I get a one hex map!

that's the task heh

ldunnplaymint:
Oh right, I am a bit confused then! What is my process if I want to get say the Istabul or the Vanilla map?

ldunnplaymint (approved):

PR #1028, opened 2024-02-07
Adding the new map system to the README

### LINKED ISSUE #1027: Update the README with the new local map deploy system
So to run a specific map in downstream you can use something like :
MAP=tiny make dev
MAP=quest-map make dev
MAP=default make dev

Or you can just do a standard make dev which will start up the single tile default map, and then run something like  ds apply -R -f ./contracts/src/maps/quest-map/ from a separate console window.

ldunnplaymint:
Resolves #1027

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,3 +191,29 @@ docker build -t ghcr.io/playmint/ds:latest .
 ```
 
 </details>
+
+<details>
+
+<summary>Running Local with different Map Setups</summary>
+
+By Default, running `Make Dev` will spawn a one hex sized map. In order to deploy locally with a different map, there are a few options
+
+### 1. Using Playmint's Maps
+Inside of the `ds/contracts/src/maps/` folder, you will find a few premade maps  by Playmint. In order to force one of these maps to be deployed with a `make dev`, you will need to add the MAP=$ arg to your command. Here are the currently supported MAP args in context:
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -191,3 +191,29 @@ docker build -t ghcr.io/playmint/ds:latest .
...
+
+### 2. Apply a map after deploying
+After doing a standard make dev, you can run the DS apply command and point it at one of the map folders. For example: `ds apply -R -f ./contracts/src/maps/quest-map/`
+
+### 3. Build your own map and deploy it
+First up, you will need to run a `make dev` and then visit `http://localhost:3000/tile-fabricator`
+
+Once in the Tile Fabricator, you can design and export a map file. 
+
+If you then rename the .yml file to a .yaml and move it into a new folder within `ds/contracts/src/maps/`, you will be able to run the ds apply command, like so:
```
chrisfarms, line 216: I think telling people to put their maps inside the repo is potentially confusing ... I think it's probably better to just talk about the the exported file directly, ie:

export the map yaml .... write it to a file mymap.yaml ... apply it with `ds apply -f ./path/to/mymap.yaml`

ldunnplaymint: Made the change

chrisfarms (approved):

PR #1030, opened 2024-02-08
Combat auto finalise

### LINKED ISSUE #999: Instantly resolve combat, after a delay
When a player initiates combat:
1. The tiles should get the red combat highlight (as they currently do) and the combat session stats are viewable.
1b. The time when the combat will happen is visible.
2. Combat hasn't begun yet. Units can still join/leave the "session", but no damage is being done
3. When the combat time is reached, the combat is instantly resolved.
- Use the current combat system. Instead of one tick per 2 secs, loop through all the ticks instantly

4. The act of starting a combat also "pays" for the resolve at the end of the combat. This means that when the combat has happened, it is immediately resolved and the game state updated.

CombatSummaryPng
Screenshot 2024-01-26 115021

hypnoshock:
# What

After combat is initiated on a building, players have a window of 30 seconds to join either side. When that time is up, combat is auto finalised by the frontend which means all the ticks of the combat are played through at once and if the attacking side is the winner, the building will be destroyed and goo distributed in bags as before. 

Currently if the attackers leave the session before the countdown is over the session will *still remain open* allowing players to join again. This could be changed if we would rather the combat session ended if there are zero attackers.

Combat start time is shown in both the summary and the modal:

image

image

# Details

We are still counting people in and out of combat when they move on and off tiles but instead of this being stored in an annotation we are connecting either a `CombatAttacker` or `CombatDefender` edge from the session to the Unit/Building. This is because there isn't any way of knowing who is standing on a particular tile when it comes to finalisation. 

An optimisation to this if we don't want to log who is present at the combat session is perhaps to incremement/decrement the number of units currently stood at a tile and submit the list of participants as calldata when we finalise. We can then check on-chain if a) The number of participants matches the unit count of a tile and b) if each of the submitted participants are truly at the combat location. The way combat quest tasks work, we might need to keep a record of combatants anyway.

## Finalisation

No done automatically by the frontend when the start blocktime set on the session has been reached and the isFinalised flag is unset. With several people playing the game at once, all the clients will attempt to finalise all finished unfinalised combat sessions and it will be the first to submit the tx who will actually finalise the session.

# Known problems

The combat quest task hasn't been refactored properly yet. At the momement any combat quest tasks will auto complete after they have been accepted. There will be a task to refactor this quest task so that it checks through all finialised sessions looking to see if the player's unit was on the attacker side and if the attackers won. We don't log the stats of the players which means we cannot similate the combat to find out the winner so we either need to:
- Log the stats for each of the participants at finalisation time
- Log the result of the combat (attackers won / defenders won) if we don't care about the details

fixes #999


ldunnplaymint:
Looking good, I did as many combos of multiple units starting/joining/leaving combat I could think of. It has made me want a countdown over the combat and a win % chance! 

Non-blocking feedback:
- I think combat should end if all attackers leave, just feels clearer
- The red highlight around an attack target makes sense to show me where I can attack from but I think it should stop appearing once I am attacking (Or only be visible in the attack intent)

ldunnplaymint (approved):

hypnoshock:
> * I think combat should end if all attackers leave, just feels clearer

Agreed. I've made it so

frontend/src/components/views/shell/index.tsx
```diff
@@ -233,6 +260,29 @@ export const Shell: FunctionComponent<ShellProps> = () => {
...
+            (s) =>
+                ({
+                    name: 'FINALISE_COMBAT',
+                    args: [s.id],
+                } as CogAction)
+        );
+
+        if (finaliseActions.length > 0) {
+            // FIXME: Would this lock up combat entirely if one of the combats couldn't resolve for some reason?
+            player.dispatchAndWait(...finaliseActions).catch((err) => console.warn(err));
```
chrisfarms, line 282: yeah the comment raises a good point, if any one of those dispatches is bad, then the whole tx will keep failing over and over.

maybe seperate dispatches would be better?  ... might be more noisey and slower, but at least it won't get stuck

Promise.all( finaliseActions.map(a => player.dispatchAndWait(a)) ).catch(...)   ?

chrisfarms (approved):

frontend/src/components/views/shell/index.tsx
hypnoshock: Agreed. Now dispatching one by one

PR #1032, opened 2024-02-09
BasicFatory.sol exported from fabricator matches example code

5p0rt5BEArD:
### What 
The building fabricator exports a hardcoded string of solidity code that should be an exact match of the examples/BasicFactory.sol

This got out of sync in #1031 

(also matched syntax for override function in commented out code - just cosmetic consistency)


hypnoshock (approved):
Nice save! 
I wonder if there is a way we can import the `BasicFactory.sol` file in as a text document instead of having to use this string..🤔

chrisfarms:
> Nice save! I wonder if there is a way we can import the `BasicFactory.sol` file in as a text document instead of having to use this string..🤔

it _should_ be possible by getting the bundler to use a file-loader so that `import thing from 'thing.sol'; magically turns it into a string ..... but I gave up trying to get it to work ... we should def try again tho if we are keeping this as it's got out of sync many times

PR #1034, opened 2024-02-12
race condition on deploy before anvil ready

chrisfarms:
we are starting to see intermittent failures on startup during the deployment of fresh contracts.

it's not clear yet why this is, but adding in a horrible sleep to both devstartup and the container startup to workaround it for the moment.

fixes: #1042  (hopefully 🤞 )


hypnoshock (approved):
I have run this > 10 times and I haven't seen the error.

PR #1036, opened 2024-02-12
Location hook

hypnoshock:
# What

Two new hooks added to base `BuildingKind` contract - `onUnitArrive` and `onUnitLeave`

I've added an example in `example-plugins` which checks for an `ID Card` in the inventory of a unit and will revert if they arrive on the building's tile if they don't have one. By reverting the move action will not complete therefore blocking their movement.

# Details

These hooks are both implemented within the `MovementRule` and are pretty self explanitory.

# Known problem

If the `onUnitArrive` is used to implement a gate, the frontend doesn't know if the criteria has been met to allow them to pass therefore we cannot show a red 'you cannot go here' tile as we're currently relying on the movement reverting to block them. In my example I'm at least displaying a message stating 'you do not have an ID card' but you cannot see this unless you click the building directly, it doesn't feed into the path finding system

If the building implementation's `onUnitLeave` reverts, the unit will not be able to leave the tile! (unless they teleport using the cli). I think having a hook for leaving is more useful than not having one so I think we should keep it. Bug or Feature if we want to implement a prison!

### DIFF
--- a/cli/test/examples/Building.sol
+++ b/cli/test/examples/Building.sol
@@ -4,7 +4,10 @@ pragma solidity ^0.8.13;
 import "@ds/IDownstream.sol";
 
 contract HammerFactory {
-    function use(Game ds, bytes24 buildingInstance, bytes24, /*mobileUnit*/ bytes calldata /*payload*/ ) public {
+    function use(Game ds, bytes24 buildingInstance, bytes24, /*mobileUnit*/ bytes calldata /*payload*/ )
+        public
+        override
+    {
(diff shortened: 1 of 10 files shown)


chrisfarms (approved):
heh yeah not so sure about the onLeave ... but it should lead to some fun at least!

do we need to check if the `srcTile != dstTile` before calling hooks? ... I haven't given it much thought, but might be a footgun around there, since there's no guarentee that prev/next locations are actually different tiles 

all seems reasonable, example is helpful 👍

PR #1037, opened 2024-02-13
Performance - Increase Load (chaos-units placing buildings)

### LINKED ISSUE #1002: more load for the performance load test map
in #994 we added a way to spin up a generated map to use as a base for checking load/performance with help from the `ds test chaos-unit`

we would like to see a bit more "load" in the form of more buildings on the map to gain more condifence in how much data we are likely to see over time .... potentially by getting the chaos-units to randomly construct a building while they are bouncing around.

Steedie:
Resolves #1002 

## What
This PR adds the `-buildings` (`-b`) option to the `concurrency`/`chaos-unit` command which expects a list of BuildingKind IDs (For example: `-b 0xbe92755c0000000000000000546391e80000000000000003,0xbe92755c0000000000000000444749c70000000000000003`, which is the Weak Duck and Weak Burger building IDs).

## [Why](https://github.com/playmint/ds/issues/1002#issue-2107485231)
We would like to see a bit more "load" in the form of more buildings on the map to gain more confidence in how much data we are likely to see over time.

## How
In the code that controls the chaos-units in `test.ts`, I have added logic to select and place a random building from the provided list:
```ts
if (buildings?.length > 0){
                let selectedBuilding = buildings[Math.floor(Math.random() * buildings.length)];
```
if there is no building on the target tile already
```ts
                const targetTileID = CompoundKeyEncoder.encodeInt16(NodeSelectors.Tile, 0, q, r, s);
                const world = await getWorld(ctx);
                const buildingOnTargetTile = getBuildingOnTile(world.buildings, targetTileID);
                if (buildingOnTargetTile === undefined){
                    await player.dispatch({ name: 'DEV_SPAWN_BUILDING', args: [selectedBuilding, q, r, s] }).then(res => res.wait());
                }
```

### To test this
Run
```make
make dev MAP=performance-test ARENAS=1
```
And then
```make
make cli && ds -n local -k 0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff9524d487fb5d57a8 test concurrency -c 1 -b 0xbe92755c0000000000000000546391e80000000000000003,0xbe92755c0000000000000000444749c70000000000000003
```
(You can change the provided BuildingKind IDs to anything deployed in your Hexwood. In this case, it's the Weak Duck and Weak Burger buildings that are created via the performance-test).

### DIFF
--- a/cli/src/commands/test.ts
+++ b/cli/src/commands/test.ts
@@ -90,7 +90,7 @@ const chaosUnit = {
             await sleep(Math.floor(Math.random() * 1000)); // jitter
             // place building
             if (buildings?.length > 0){
-                let selectedBuilding = buildings[Math.floor(Math.random() * buildings.length)];
+                const selectedBuilding = buildings[Math.floor(Math.random() * buildings.length)];
                 const targetTileID = CompoundKeyEncoder.encodeInt16(NodeSelectors.Tile, 0, q, r, s);
                 const world = await getWorld(ctx);
                 const buildingOnTargetTile = getBuildingOnTile(world.buildings, targetTileID);
--- a/cli/src/commands/test.ts
+++ b/cli/src/commands/test.ts
@@ -1,11 +1,12 @@
 import { ethers } from 'ethers';
 import { pipe, take, toPromise } from 'wonka';
-import { CompoundKeyEncoder, NodeSelectors, getCoords } from "@downstream/core";
+import { CompoundKeyEncoder, NodeSelectors, getCoords, WorldTileFragment } from "@downstream/core";
+import { getBuildingAtTile } from '@downstream/core/src/utils';
 import util from 'node:util';
 import { spawn } from 'node:child_process';
+import { getWorld } from './get';
 
 const spawnAsync = util.promisify(spawn);
-let buildings: string[];
 
 const concurrency = {
     command: 'concurrency',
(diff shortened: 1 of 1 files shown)


cli/src/commands/test.ts
```diff
@@ -30,11 +37,15 @@ const concurrency = {
         const procs: ReturnType<typeof spawnAsync>[] = [];
         for (let wallet of wallets) {
             console.log(`starting chaos-unit for ${wallet.privateKey}`);
-            procs.push(spawnAsync('ds', [
+            let args = [
```
chrisfarms, line 40: does this need to be a `let`?

we should try to use `const` wherever possible to help our linters catch weird issues.

cli/src/commands/test.ts
```diff
@@ -67,11 +87,27 @@ const chaosUnit = {
...
-            await sleep(Math.floor(Math.random() * 2000)); // jitter
+            await sleep(Math.floor(Math.random() * 1000)); // jitter
+            // place building
+            if (buildings?.length > 0){
+                let selectedBuilding = buildings[Math.floor(Math.random() * buildings.length)];
+                const targetTileID = CompoundKeyEncoder.encodeInt16(NodeSelectors.Tile, 0, q, r, s);
+                const world = await getWorld(ctx);
+                const buildingOnTargetTile = getBuildingOnTile(world.buildings, targetTileID);
+                // console.log(`target tile (${targetTileID}) building: ${buildingOnTargetTile}`);
+                if (buildingOnTargetTile === undefined){
```
chrisfarms, line 98:
```suggestion
                if (typeof buildingOnTargetTile === 'undefined'){
```

javascript is dumb and allows you to do....

```
globalThis.undefined = "lol";
"lol" == undefined // true
```

... so it's generally best to avoid directly checking equality of undefined.

chrisfarms (approved):
LGTM .. minor non-blocking comments

Steedie:
Thanks for the feedback and info. I've made both these changes. Testing now then will push again.

Issue #1008, opened 2024-02-02
Ability to deploy the performance test to hexwood instances


ldunnplaymint:
related to #1007 , we want to be able to provision instances with the performance-map ... but this map additionally needs an option for a size to generate so cannot be selected like a normal map from the list

PR #1038, opened 2024-02-13
allow manually applying the generated "performance-test" map

chrisfarms:
### what

we want to be able to manually deploy the generated performance map to remote instances

which means separating the "generation" away from the devstartup script

👀  using the performance map is now a two stage process:

First you run:

```
NUM_ARENAS=4 make contracts/src/maps/performance-test
```

.... this populates the `contracts/src/maps/performance-test` dir with all the generated manifests.

Then we can treat it like any other map dir:

```
MAP=performance-test make dev
```

This means that we can also now apply the map to a remote instance via:

```
ds apply -n hexwoodX -R -f contracts/src/maps/performance-test
```

Use `make clean` to destroy the generated map and build another one

### notes

this is a bit more long-winded that before (as the dev script no longer detects it needs to generate the missing map), but since we don't expect to do it very often I think that's probably ok.

resolves: #1008


Steedie:
Good stuff! I've just run through this process and it's working on my end. The code changes look good and make sense. ✅ from me.

Steedie (approved):
Good stuff! I've just run through this process and it's working on my end. The code changes look good and make sense. ✅ from me.

Issue #1029, opened 2024-02-08
change the docker compose setup to use prebuilt images by default


chrisfarms:
...so that

* people can spin up locally without needing unity etc

PR #1040, opened 2024-02-13
use prebuilt images in default docker compose config

chrisfarms:
### what

disables the build steps in docker compose config and instead points to the `hexwood0` tags for the last built images

### why 

we don't use docker compose for dev, but it is useful if you want a local copy of the game running without needing all the build tooling, unity, and dependencies

resolves: #1029 

### testing

please see the updated README and try it out to check it does actually let you pull down the images .... it will run/startup pretty slow for those of you on M1 macs, since we are only building amd64 images at this point... but it should be ok once it's finally up

### DIFF
--- a/.env
+++ b/.env
@@ -1,0 +1,1 @@
+MAP=tiny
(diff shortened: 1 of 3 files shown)


PlaymintPaul (approved):
Works in Windows

PR #1043, opened 2024-02-14
Hexcraft integration

hypnoshock:
# What

Added hexcraft map and all buildings to our `contracts/src/maps` folder so it can be deployed with `MAP=hexcraft`
I have also added `hexcraft` to the list of maps in the github actions so it can be deployed to a hexwood

NOTE: `arena-1kx` has been effectively renamed to `hexcraft`. If this wasn't what I was supposed to do, I can rename it back

# Bug Fixes

I found two bugs with combat frontend:
- The world list of mobile units wasn't been sent to the combat action panel which meant the only unit to be entered into combat was the one that instigated the combat instead of all the units stood on the tile. This solves an issue that we've been having with Downstream for MONTHS. There was a report of units not auto joining combat and this was the reason.
- The life of a building wasn't having the 10x multiplier applied therefore the reported life was incorrect

# Testing
- I have tested that hexcraft buildings can only be crafted if a crafting hammer is in the inventory
- crafting hammer is taken out of inventory after building craft
- I have tested that only one class of item can be crafted i.e. the HQ is tracking which item a player crafted for their unit
- Tested joining and starting a game
- Tested trying to join a game after it has started (The revert message stating that you cannot join is shown in the console)
- Tested destroying a tower
  - The HQ gives a warning that a particular team's tower is missing which I think is the end state of the game.


chrisfarms (approved):
good combat bug squishing 💪 

not-for-this-pr: what are you using for YAML formatting? ... be good if we can get the formatting settings in the repo

hypnoshock:
> good combat bug squishing 💪
> 
> not-for-this-pr: what are you using for YAML formatting? ... be good if we can get the formatting settings in the repo

Sorry I didn't notice the comment. It's prettier that's doing it and in the prettier output tab I'm seeing this after saving a yaml file:

```
["INFO" - 15:33:21] Detected local configuration (i.e. .prettierrc or .editorconfig), VS Code configuration will not be used
["INFO" - 15:33:21] Prettier Options:
{
  "filepath": "/Users/hypnoshock/Projects/playMint/ds/contracts/src/maps/default/map.yaml",
  "parser": "yaml",
  "useTabs": false,
  "tabWidth": 2,
  "endOfLine": "lf"
}
["INFO" - 15:33:21] Formatting completed in 16ms.

```

I'm not sure which .prettier.rc it's referring to, I think it might be referring to the one in the base of frontend but it IS using some defaults. If we wanted consistency we could add the settings quoted above to a .prettier.rc in the root of the repo I think

5p0rt5BEArD (approved):
Brief local test all looks good

PR #1044, opened 2024-02-14
Unit Team Distinction (Change color of a unit from plugin)

### LINKED ISSUE #1023: As a team based experience maker I want to show which team each Unit is on so players can see whats going on
Would be useful for all open croissant games

Tonk Attack
Hexcraft
Ducks v Burger

Steedie:
## What
Changes to Unity and the frontend to allow a mobile unit's color to be changed from plugins.

### Unity diorama
image

### Custom plugin changing unit's color
image

## [Why](https://github.com/orgs/playmint/projects/1/views/1?pane=issue&itemId=52368565)
This has been a common request during our playtests of Duck Burger, Hexcraft, and more.
_"As a team based experience maker I want to show which team each Unit is on so players can see whats going on"_

## How
First, changes in Unity to MobileUnitData, MobileUnitController, and MobileUnitDiorama, resulting in the diorama screenshot above.
Then changes to the frontend, mostly following the format of how plugins can change tile colours.

## Info
To test this in a local build, I created a very simple custom plugin and hard-coded my unit's ID into the map:
```js
    const mapObj = [];
    mapObj.push(
        {
            type: "unit",
            key: "color",
            id: "0x0945e868000000000000000000000000000000000000033b",
            value: "#ff0000",
        }
    )

    return {
        version: 1,
        map: mapObj, ....
```

edit: team colours have now been implemented into the DuckBurgerHQ example building.

Resolves #1023

### DIFF
--- a/frontend/src/components/map/MobileUnit.tsx
+++ b/frontend/src/components/map/MobileUnit.tsx
@@ -146,7 +146,7 @@ export const MobileUnits = memo(
         selectedMobileUnitID,
         onClickMobileUnit,
         playerID,
-        randomUnitProperties,
+        pluginProperties,
     }: {
         currentBlock: number;
         mobileUnits?: WorldMobileUnitFragment[];
@@ -154,7 +154,7 @@ export const MobileUnits = memo(
         selectedMobileUnitID?: string;
         playerID?: string;
         onClickMobileUnit: (id: string) => void;
-        randomUnitProperties: PluginMapProperty[];
+        pluginProperties: PluginMapProperty[];
     }) => {
         const [unitPositions, setUnitPositions] = useState({}); // New state for positions
 
@@ -202,7 +202,7 @@ export const MobileUnits = memo(
             () =>
                 units.map((u) => {
                     // Make this like Tile.tsx line 71
-                    const color = randomUnitProperties.find((prop) => prop.id == u.id)?.value.toString();
+                    const color = pluginProperties.find((prop) => prop.id == u.id)?.value.toString();
                     return (
                         <MobileUnit
                             sendScreenPosition={true}
@@ -221,7 +221,7 @@ export const MobileUnits = memo(
                         />
                     );
                 }),
-            [onClickMobileUnit, selectedMobileUnitID, units, updatePosition, randomUnitProperties]
+            [onClickMobileUnit, selectedMobileUnitID, units, updatePosition, pluginProperties]
         );
 
         const unitIcons = useMemo(
--- a/frontend/src/components/map/MobileUnit.tsx
+++ b/frontend/src/components/map/MobileUnit.tsx
@@ -1,6 +1,6 @@
 import { getTileHeightFromCoords } from '@app/helpers/tile';
 import { UnityComponentProps, useUnityComponentManager } from '@app/hooks/use-unity-component-manager';
-import { WorldBuildingFragment, WorldMobileUnitFragment, getCoords } from '@downstream/core';
+import { WorldBuildingFragment, WorldMobileUnitFragment, getCoords, PluginMapProperty } from '@downstream/core';
 import { memo, useCallback, useMemo, useState } from 'react';
 import Icon from './Icon';
 import { getBuildingAtTile } from '@downstream/core/src/utils';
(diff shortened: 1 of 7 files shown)


Steedie:
Having trouble running csharpier on Windows CMD:
image
Can't remember how I did it last time

chrisfarms:
> Having trouble running csharpier on Windows CMD:

try:

```
cd map
dotnet tool run dotnet-csharpier .
```

frontend/src/components/map/MobileUnit.tsx
```diff
@@ -142,13 +146,15 @@ export const MobileUnits = memo(
...
         playerID,
+        randomUnitProperties,
     }: {
         currentBlock: number;
         mobileUnits?: WorldMobileUnitFragment[];
         buildings: WorldBuildingFragment[];
         selectedMobileUnitID?: string;
         playerID?: string;
         onClickMobileUnit: (id: string) => void;
+        randomUnitProperties: PluginMapProperty[];
```
chrisfarms, line 157: hehe as much as it makes me laugh that the `randomProperties` from the initial test example has created a new standard, we should probably name this more sensibly to avoid confusion.

maybe `pluginProperties` or something like that.

map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
```diff
@@ -109,6 +111,13 @@ protected void Update()
             }
         }
 
+        // Body color
+        if (_nextData.color?.Length > 0 && (_prevData?.color ?? "") != _nextData.color)
+        {
+            Color newColor = RGBStringToColor(_nextData.color);
```
chrisfarms, line 117: does this need to be a special case of only accepting RGB values?

it might be less of a footgun if the `color` property on units, worked the same way that it works on other components where it accepts CSS color values.

Example of using the ColorUtility here: https://github.com/playmint/ds/blob/main/map/Assets/Scripts/Components/Tile/TileController.cs#L45

chrisfarms (changes requested):

map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
Steedie: Ah just what I needed! This is now implemented:
```cs
            string colorString = _nextData.color;
            if (!string.IsNullOrEmpty(colorString) && !colorString.StartsWith("#"))
            {
                colorString = "#" + colorString;
            }
            else if (string.IsNullOrEmpty(colorString))
            {
                colorString = "#" + _defaultBodyColor;
            }

            ColorUtility.TryParseHtmlString(colorString, out Color targetColor);
            renderers[0].material.color = targetColor;
```

map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
```diff
@@ -109,6 +112,23 @@ protected void Update()
             }
         }
 
+        // Body color
+        if ((_prevData?.color ?? "") != _nextData.color)
+        {
+            string colorString = _nextData.color;
+            if (!string.IsNullOrEmpty(colorString) && !colorString.StartsWith("#"))
```
chrisfarms, line 119: non-blocking: I _believe_ TryParseHtmlString handles more than just `#` prefixed color strings ... for example `red` and `rgb(0,0,0)` are both valid ... so it looks like this bit where it forces the `#` prefix might break those things, if I'm reading it right. .... non-blocking since I think in reality we do use the hex values, so it's probably a non-issue

chrisfarms (approved):

map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
Steedie: Good point. I've now modified this so it works according to Unity's [ColorUtility.TryParseHtmlString documentation](ColorUtility](https://docs.unity3d.com/ScriptReference/ColorUtility.html).TryParseHtmlString)

These are examples of inputs that now work: "#f00", "#ff0000", "f000", "ff000000", and "red".
"ff0000" no longer works, but allows inputs like "red", and "orange" to also work.

PR #1059, opened 2024-02-19
Only set team colours when in range (DuckBurgerHQ)

Steedie:
## What
Requires the unit to be in range of the DuckBurgerHQ before setting the unit's colour.

In the image below, it shows 2 units who are both on a team (from the DuckBurgerHQ on the left). One is standing outside the range of 5 tiles.
image

## Why
Because a player might have joined a game of Duck Burger, but wondered off to another game before Duck Burger was reset.

## How
It checks the distance between the unit and the DuckBurgerHQ before assigning the colour to make sure it's in range:
```js
if (distance(getTileCoords(unitCoords), getTileCoords(hqCoords)) > 5) {
            continue;
        }
```


hypnoshock (approved):
Looks good to me

PR #1060, opened 2024-02-19
allow spinning up "tonk" map with docker compose

chrisfarms:
### What

* Adds a basic tonk map to contracts/src/maps
* Updates the docker compose config to allow spinning up the game + tonk-services for testing
* Adds a new `ds.config` property available to plugins that exposes [the game's config](https://github.com/playmint/ds/blob/tonk-local/frontend/public/config.json) to allow tonk plugins to discover correct endpoint `ds.config.tonkEndpoint`
* detect if there is a `postinstall.js` file in the map folder, and execute it during deploy ... which is then used to register tonk buildings
* warn if you try to do `MAP=tonk make dev` since you would need the services built manually yourself
* mounts the maps dir into the docker compose container ... so you can alter maps without needing to rebuild the image

### Why

This is all the bits we need to be able to dynamically setup a tonk-attack game ...  ~~in a followup PR we will update the Chart to make use of all of these bits to include tonk-services during deployment~~ (chart manifests now included in this PR)

### Testing notes

Since this PR modifies the frontend + contracts containers, you will need to pass `--build` to test ... for example:

```
MAP=tonk docker compose --profile=tonk up --build
```

### DIFF
--- a/.devstartup.js
+++ b/.devstartup.js
@@ -13,6 +13,17 @@ const DEPLOYER_PRIVATE_KEY = "0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff
 const MAP = process.env.MAP || "default";
 
 async function handleStartup(){
+    // bail if MAP=tonk
+    if (MAP === 'tonk') {
+        console.error('');
+        console.error('MAP=tonk is not supported via `make dev`');
+        console.error('');
+        console.error('Either use: docker compose --profile tonk up');
+        console.error('');
+        console.error('Or use MAP=default and then work out how to build and run tonk services manually (requires rust toolchain)');
+        console.error('');
+        process.exit(1);
+    }
     // which map to deploy (relative to contracts)
     const mapPath = path.join('./src/maps/', MAP);
     if (!fs.existsSync(path.join('contracts', mapPath))) {
--- a/.devstartup.js
+++ b/.devstartup.js
@@ -18,7 +18,7 @@ async function handleStartup(){
         console.error('');
         console.error('MAP=tonk is not supported via `make dev`');
         console.error('');
-        console.error('Either use: docker compose --profile tonk up');
+        console.error('Either use: `MAP=tonk docker compose --profile tonk up`');
         console.error('');
         console.error('Or use MAP=default and then work out how to build and run tonk services manually (requires rust toolchain)');
         console.error('');
--- a/contracts/src/maps/tonk/postinstall.js
+++ b/contracts/src/maps/tonk/postinstall.js
@@ -1,0 +1,95 @@
+const TONK_URL_HTTP = process.env.TONK_URL_HTTP;
+if (!TONK_URL_HTTP) {
+    throw new Error("TONK_URL_HTTP env var not set");
+}
+
+async function register_building(id, readable_id, is_tower, message, location) {
+    var myHeaders = new Headers();
+    myHeaders.append("Content-Type", "application/json");
+
+    var raw = JSON.stringify({
+        id: id,
+        is_tower: is_tower,
+        task_message: message,
+        readable_id: readable_id,
+        location,
+    });
+
+    var requestOptions = {
+        method: "POST",
+        headers: myHeaders,
+        body: raw,
+        redirect: "follow",
+    };
+
+    return fetch(`${TONK_URL_HTTP}/building`, requestOptions)
+        .then((response) => response.text())
+        .then((result) => console.log(result))
+        .catch((error) => console.log("error", error));
+}
+
+const HEX_DUMP_MESSAGE = "Download garbage data";
+const MEME_GEN_MESSAGE = "Download steamed hams";
+const SELFIE_POINT_MESSAGE = "Download vacation selfies";
+
+// async function register_all() {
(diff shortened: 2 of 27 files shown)


contracts/src/maps/tonk/postinstall.js
```diff
@@ -0,0 +1,95 @@
...
+        .then((response) => response.text())
+        .then((result) => console.log(result))
+        .catch((error) => console.log("error", error));
+}
+
+const HEX_DUMP_MESSAGE = "Download garbage data";
+const MEME_GEN_MESSAGE = "Download steamed hams";
+const SELFIE_POINT_MESSAGE = "Download vacation selfies";
+
+// async function register_all() {
```
5p0rt5BEArD, line 35: A comment on why this version is commented out?

contracts/src/maps/tonk/postinstall.js
```diff
@@ -0,0 +1,95 @@
...
+//         '0x0', '0x05', '0x02', '0xfff9'
+//     ])
+//     await register_building("0x34cf8a7e0000000000000000000000000000fff400030009",false, BREAKPOINT_MESSAGE, [
+//         '0x0', '0xfff4', '0x03', '0x09'
+//     ])
+// }
+
+async function register_all() {
+    await register_building(
+        "0x34cf8a7e000000000000000000000000000000010000ffff",
```
5p0rt5BEArD, line 66: A comment on how these ids are derived? Especially useful is this is now first class example code

hypnoshock (approved):
Looks good to me however I think 2 of the hex coords in the postInstall script are wrong.

chrisfarms:
thanks @hypnoshock  ... I guess I picked the wrong one of the commented out bits

hypnoshock:
Also I tried

hypnoshock:
> thanks @hypnoshock ... I guess I picked the wrong one of the commented out bits

I don't think any of them quite fit TBF

.devstartup.js
```diff
@@ -13,6 +13,17 @@ const DEPLOYER_PRIVATE_KEY = "0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff
 const MAP = process.env.MAP || "default";
 
 async function handleStartup(){
+    // bail if MAP=tonk
+    if (MAP === 'tonk') {
+        console.error('');
+        console.error('MAP=tonk is not supported via `make dev`');
+        console.error('');
+        console.error('Either use: docker compose --profile tonk up');
```
5p0rt5BEArD, line 21: add `--build` to suggested docker command

5p0rt5BEArD: should also note that MAP=tonk needs setting in the .env file

5p0rt5BEArD (changes requested):
Tested and played a couple of rounds of 1v1 locally and all features seem to work
Minor comment comments

.devstartup.js
chrisfarms: you should not need to add `--build` ... you only need it right now because the `hexwood0` image does not contain the changes in this PR yet (it will once it's merged)

contracts/src/maps/tonk/postinstall.js
chrisfarms: I didn't write any of this it's all from the tonk repo ... I can tidy it up, but it might make copying in any future changes harder 🤷‍♂️

5p0rt5BEArD: yeah fair enough - if its copied straight from the source then leave it - don't want to do a comment/edit pass on all their stuff

contracts/src/maps/tonk/postinstall.js
chrisfarms: I honestly don't know what anything in this file really does, I just know I have to run it heh ... I guess they are building ids somehow dug out of the graphql query by hand

it's the "registration" script from tonk's repo ... as per other comment, gonna keep as is, as it's not really ours

.devstartup.js
chrisfarms: ah yeah added MAP=tonk to the docker line thanks... if you are seeing this warning, then it implies you have been able to pass an environment variable so no need to edit the .env file

contracts/src/maps/tonk/postinstall.js
```diff
@@ -0,0 +1,95 @@
...
+    );
+    await register_building(
+        "0x34cf8a7e000000000000000000000000000000050002fff9",
+        "DATA DUMP EAST",
+        false,
+        MEME_GEN_MESSAGE,
+        ["0x0", "0x05", "0x02", "0xfff9"],
+    );
+    await register_building(
+        "0x34cf8a7e0000000000000000000000000000fffbfffe0007",
```
hypnoshock, line 87:
```suggestion
        "0x34cf8a7e0000000000000000000000000000fffcfffd0007",
```

contracts/src/maps/tonk/postinstall.js
hypnoshock: Also I just realised I didn't suggest to update the IDs to match the coords!! 😱

5p0rt5BEArD (approved):

PR #1061, opened 2024-02-19
Clientside blocker building

hypnoshock:
# What

We can now use plugin code to supply the frontend with blocker tiles by using `PluginMapProperty` objects with type `tile` and the new key of `blocker`

I have updated the gate example building to be a full map example which lives in `contracts/src/maps/example-gate`

In the example the unit must collect a key item from the bag to unlock the gate building. I am also colouring the tile to show the lock state of the tile. This is optional i.e. the tile being a blocker doesn't intrinsically colour it.

Before collecting key:
image
image

After Collecting key:
image

# Details

The gist of how I got this to work is that: 

- I've stopped the frontend assuming that all `PluginMapProperty` objects of type tile change the colour and have added the key `blocker`
- The value for the new key `blocker` can be either the string value `true` or `false`
- I'm passing these `PluginMapProperty` objects into the action panel and all the containing actions
- I've updated the path finding and the move code to check if a tile `isBlockerTile` by checking the list of `PluginMapProperty` for any objects with key `blocker` with a value of `true`


contracts/src/maps/example-gate/Gate/Gate.sol
```diff
@@ -16,10 +15,14 @@ contract ExampleGate is BuildingKind {
         // Example of logging the last unit to arrive at a building
         _setDataOnBuilding(ds.getDispatcher(), buildingInstanceID, "lastUnitToArrive", bytes32(mobileUnitID));
 
-        // Example of not allowing a unit to stand on the building tile unless they have the ID Card item
-        ( /*uint8 bagSlot*/ , /*uint8 itemSlot*/, uint64 balance) =
-            _findItem(ds.getState(), mobileUnitID, ItemUtils.IDCard());
-        require(balance > 0, "ExampleGate: Unit does not have ID Card");
+        // Example of not allowing a unit to stand on the building tile unless they have the Gate Key item
+
+        // NOTE: There's a problem with generating the ID using Node.Item as it doesn't match the one generated by the cli tool.
```
chrisfarms, line 20: 🤔 guess we missed something when we moved to unique item names .... let's add an issue to track this

frontend/src/components/map/Tile.tsx
```diff
@@ -68,7 +68,9 @@ export const Tiles = memo(
         const tileComponents = useMemo(
             () =>
                 (tiles || []).map((t) => {
-                    const color = randomTileProperties.find((prop) => prop.id == t.id)?.value.toString();
+                    const color = randomTileProperties
```
chrisfarms, line 71: **nit:** I know it's not your code, but can we rename `randomTileProperties` to `pluginTileProperties` or something like that now that this code is sticking around properly ... that code was originally just an example that has overstayed it's welcome hence the silly name 🙏

frontend/src/helpers/pathfinding.ts
```diff
@@ -35,7 +37,10 @@ export function getPath(
         // ignore tiles with blockers on them
         const building = getBuildingAtTile(buildings, t.tile);
         const key = `${q}:${r}:${s}`;
-        if (building && getBuildingCategory(building?.kind) === BuildingCategory.BLOCKER) {
+        if (
+            isBlockerTile(t.tile, tilesModifiedByPlugins) ||
```
chrisfarms, line 41: **non-blocking:** i think it's probably ok because the list of `tilesModifiedByPlugins` is _probably_ a small list .... but this is in a very hot loop ... it might be worth precomputing the set of affected tile ids outside of the `for` loop so you can do a fast `has/in` check here

chrisfarms (approved):
what a lovely PR ... it has pictures, nice walkthrough, examples, and fun feature ... 10/10 would review again

minor nit and a comment, leave it up to you if you think they are worth addressing

frontend/src/components/map/Tile.tsx
hypnoshock: I should have done that when I saw it as I did notice it was weird! Fixed also I've changed where I was calling it something like `tilesModifiedByPlugins` to also be `pluginTileProperties` so it matches

frontend/src/helpers/pathfinding.ts
hypnoshock: I'm now adding them all to the blocker map outside of the for loop

```
    // add the dynamic blockers to blocker map
    pluginTileProperties.forEach((p) => {
        if (p.key == 'blocker' && p.value === 'true') {
            const [q, r, s] = getTileCoordsFromId(p.id);
            const key = `${q}:${r}:${s}`;
            blockerMap.set(key, true);
        }
    });
```

PR #1069, opened 2024-02-21
pull=always

chrisfarms:
add --pull=always to the compose command example ... since we move too fast to assume old images will keep working


5p0rt5BEArD (approved):
I think we should have some instructions on iterating with the docker deploy as per discord chat - I'll do that a separate pr though

PR #1073, opened 2024-02-21
Multi game map

JackPlaymint:
This PR adds a new map called croissant, which contains an area for Tonk, Hexcraft and Duck vs Burger.

I had to do a fair bit of folder structure rearranging in the contracts/src subfolders to get it to play nicely with the symlinks.
The main problem being that the recursive argument on ds apply stops one folder in when looking at symlinked folders, so each folder containing relevant yaml had to be symlinked individually rather than via a nice parent folder (as was the original plan).
There was also a bit of trouble with relative file paths within the yaml itself, specifically DvB's TeamBuildings.yaml had some references to BasicFactory, but the links would no longer work. To get around this I just added a copy of the BasicFactory scripts directly to the DvB folder.
Finally there was an issue with the docker image looking at a past version of anything outside of src/maps, which caused issues with changes in src/example-plugins.
Changing the docker-compose.yaml to `volumes:  - ./contracts/src:/contracts/src` rather than `volumes: - ./contracts/src:/contracts/src/maps` seemed to resolve that issue.

I tested through each game as best I could on my own and they seem to be functioning well, if a bit slow.


.env
```diff
@@ -1 +1 @@
-MAP=tiny
```
chrisfarms, line 1: can we keep this as `tiny` by default 🙏

5p0rt5BEArD:
@chrisfarms does something need to be done to make this map build with the tonk docker profile? I can't see where the git workflow sets the profile.

chrisfarms:
> @chrisfarms does something need to be done to make this map build with the tonk docker profile? I can't see where the git workflow sets the profile.

yes ... [this line](https://github.com/playmint/ds/blob/main/chart/templates/tonk.yaml#L1) will need updating to check for both "tonk" and "croissant"

the templating language is a probably not very intuitive if you haven't used it before ... ([see here for info](https://helm.sh/docs/chart_template_guide/function_list/))

I think it'll be something like:

```
{{ if or (eq $.Values.map "tonk") (eq $.Values.map "croissant") }}
```

5p0rt5BEArD:
I'm concerned about the maps directories becoming confusion so have tried to work around some of the issues arising from symlinks and removed some unnecessary code:

- Display Buildings don't need to run any code so have removed their implementation, plugin and always active settings
- DuckBurger folder now has symlinks to the BasicFactory code files rather than duplicating
- The croissant map symlinks to the top level folder only for tonk and hexcraft - this is possible by adding the "follow" flag to the glob command used by the cli code, which tells glob to recursively follow symlinked folders when searching.

5p0rt5BEArD:
This has been tested with the team a couple of times. There are some performance issues with the croissant map but nothing that is introduced by this PR so its good to go,

PR #1074, opened 2024-02-21
Tonk Attack: Role (Team) Colours

### LINKED ISSUE #1052: Assign team colours for Tonk Attack

Steedie:
## What
"Sentient" players now see all players as blue, and players who are "Bugged" will see fellow bugged players as red, and sentient players as blue.
image

When a player is voted out, they should become a dark blue colour, or a dark red colour if you are "Bugged". (This needs slightly larger scale testing because it's possible that they're removed from the player list in which case they would become grey instead).

When the game is over, if the bugged team wins, they will be revealed to all players as red, while the losers become un-coloured.
This screenshot shows the case where the bugged ("Brainwashed") have won:
image

## Why
We want it to be more clear as to what's going on while playing Tonk Attack: Who's on my team? Who's still in the game? Who won?

## How
`alwaysActive` has been enabled so colors will be rendered at all times. 

We loop through each player to see what info we have on them:
```js
players.forEach((p) => {
        switch(p.role) {
```

Bugged units can see what role each unit has (from the client) whereas Sentient units can't, so we use this info to determine the colours that we render each unit.

## Current Limitations
We wanted to be able to see who the Bugged units are at the end if the Sentient team wins, however, this may not be possible in its current state since when a unit is voted out, they're removed from the player list.

Resolves #1052


5p0rt5BEArD (commented):
Code looks good
Tested locally with 2 players
Holding off approval until we've completed a team test

chrisfarms:
from playtest: looks like the "turns the bugged red at the end" isn't working quite right ... it either showed all the remaining people as red or showed the wrong people as red

Steedie:
## Changes
Units who have joined the game and the game state is "Lobby" are turned purple:
image

When a sentient unit is Tonked, they are eliminated and displayed as dark blue:
image

Problem: In the next round shown in the screenshot below, sentients vote out a bugged unit, but when the bugged unit is eliminated, the intended behavior is for it to turn dark red, however, the unit ID of one who was voted out does not get saved:
image

This is a screenshot of the eliminated player list. It shows the mobile_unit_id of the sentient eliminated unit, but not the bugged one.
image

In this screenshot, the bugged "brainwashed" unit has won. again, you can see that since the (sentient) unit was voted out, the unit id isn't stored in the "eliminated" list:
image

In this screenshot, the bugged "brainwashed" win. The selected unit is sentient, and the other is bugged, but here they are displayed as yellow because sentient units can't see the roles of other (still alive) units:
image


## Questions
While we don't always get access to the mobile unit ID, we do always have the mobile unit display name (but in all caps), is it possible to recalculate the unit ID using this so we can set the colour?
If not, and assuming the cases where we can't get the unit IDs won't be resolved without the tonk server code being changed, would it be better to simplify the logic to make it more consistent? For example: only seeing if a player is either in the lobby, or playing the game, as well as bugged units seeing other bugged units as red.

chrisfarms:
good stuff 👍 

> is it possible to recalculate the unit ID using this so we can set the colour?

you cannot calculate the id from the name ... however you could probably find the unit with that name:

```
    const jeffId = world?.mobileUnits.find((u) => u.name?.value === 'jeff')?.id;
```

... this of course is **not reliable**, since names are not unique, and players can change their unit names whenever they like.

it might be acceptable as a "best effort" kind of thing ...but I'm not really sure it's worth the complexity or potential confusion from someone looking at that code and thinking it's a good idea

> would it be better to simplify the logic to make it more consistent? For example: only seeing if a player is either in the lobby, or playing the game, as well as bugged units seeing other bugged units as red.

I think that is totally reasonable yes ... the primary purpose of this task was to make it clearer who is playing and who is not .... everything else was just potential icing on the cake, and if it's not practical it's fine to not do it

Steedie:
> good stuff 👍
> 
> > is it possible to recalculate the unit ID using this so we can set the colour?
> 
> you cannot calculate the id from the name ... however you could probably find the unit with that name:
> 
> ```
>     const jeffId = world?.mobileUnits.find((u) => u.name?.value === 'jeff')?.id;
> ```
> 
> ... this of course is **not reliable**, since names are not unique, and players can change their unit names whenever they like.
> 
> it might be acceptable as a "best effort" kind of thing ...but I'm not really sure it's worth the complexity or potential confusion from someone looking at that code and thinking it's a good idea
> 
> > would it be better to simplify the logic to make it more consistent? For example: only seeing if a player is either in the lobby, or playing the game, as well as bugged units seeing other bugged units as red.
> 
> I think that is totally reasonable yes ... the primary purpose of this task was to make it clearer who is playing and who is not .... everything else was just potential icing on the cake, and if it's not practical it's fine to not do it

Thanks for the info. Based on this idea, I had a look at using the player ID to get the first mobile unit of the player, but this would also break if they had old or extra units in the map.

I'll make some changes now to make it all more consistent.

Steedie:
## New updates
### Changes for consistency
- Sentients win? They see everyone as light blue.
- Sentients lost? They see everyone as dark blue.
- Brainwashed can always see everyone's role.
- When a unit is eliminated in any way, they'll become un-coloured.

5p0rt5BEArD (approved):
Latest changes make sense

PR #1076, opened 2024-02-23
cli version 0.0.19 publish

hypnoshock:
# What

published v0.0.19 of `ds-cli`
This includes the change to `schema.sol` that encodes Item IDs the same way as the applier does as in the IDs are a hash of `item/[item name]` instead of a hash of the name and recipe


5p0rt5BEArD (approved):
👍

PR #1080, opened 2024-02-23
Added Billboard building

JackPlaymint:
This PR adds a new building type which can be used to display an image from a provided URL directly on a model in the game world.
The model currently used is temporary and awaiting final art integration from Tom,  ̶b̶u̶t̶ ̶t̶h̶i̶s̶ ̶t̶a̶s̶k̶ ̶i̶s̶ ̶f̶o̶r̶ ̶p̶l̶a̶c̶e̶h̶o̶l̶d̶e̶r̶ ̶a̶r̶t̶ ̶s̶o̶ ̶I̶'̶m̶ ̶s̶h̶o̶v̶i̶n̶'̶ ̶i̶t̶ ̶i̶n̶ ̶a̶n̶y̶w̶a̶y̶!̶ Tom has made the model! So it's all in and integrated!

The building uses the 'model' parameter in the building yaml for the url string, you can find an example in the croissant map folder.

Images support clipped transparency only, so no dope fades. Also the background colour is always black (maybe it should be modifiable)?
No animation support, although that would be cool.
Also no content moderation checks or anything, so go nuts! I can't stop you.


5p0rt5BEArD:
I was expecting the image data to be loaded from react code and sent to unity as component data, rather than sending a url for unity to load via a webrequest. That would seem to fit with the pattern of Unity being a dumb renderer but webrequest is probably easier...

was this discussed in a kick off?

chrisfarms:
> was this discussed in a kick off?

we said it should pass a URL yeah.

and I was def expecting sending the URL and letting unity work out what to do with it, for a few reasons:

* serializing data to send to unity is very expensive ... and should be kept to a minimum ... a url is small, a giant unbounded blob of bytes is big
* that blob of bytes must also go into JSON ... there is no "bytes" type in JSON, so this would probably mean Base64-ing it ... which adds like 30% to the size, and requires decoding on the other end
* we also don't know what bytes unity is going to need, do we need to convert to bitmap (huge!) support multiple formats? ...add new decoders for every image type etc.

... so yeah much easier and more performant to use the exposed unity APIs for fetching and decoding imo

5p0rt5BEArD:
I think the way this has been implemented means that the image cannot be changed by plugins without changing the buildingkind model value in solidity code.

This might not have been in the issue spec but I think that it would be useful to builders to be able to change the image.

Should this follow the display building pattern and all the url to be set from `map:` plugin return block data?

contracts/src/maps/croissant/kinds/billboard.yaml
5p0rt5BEArD, line 1: I think this should go in its own maps/billboard folder, rather than as part of the croissant map.

Maybe its just a two tile map with this billboard building placed one tile away form the centre.

contracts/src/maps/croissant/kinds/billboard.yaml
```diff
@@ -0,0 +1,14 @@
+---
+kind: BuildingKind
+spec:
+  category: billboard
+  name: Basic Billboard
+  description: Shows an image
+  model: http://localhost:3000/favicon-28978_favicon-192.png
```
5p0rt5BEArD, line 7: Will this work with urls that are not localhost?

JackPlaymint: It works with some, this image was hosted somewhere on the interwebs

chrisfarms: it will work for any image host that allows it via CORS

5p0rt5BEArD:
> > I think the way this has been implemented means that the image cannot be changed by plugins without changing the buildingkind model value in solidity code.
> > Should this follow the display building pattern and allow the url to be set from `map:` plugin return block data?
> 
> I may have misunderstood something in the implementation, why wouldn't the url be updatable from a plugin?

The buildingkind.model is used for the image display.

You can set this from solidity plugin code but not javascript plugin code.

Maybe that's correct

Lets chat about it after standup tomorrow

contracts/src/maps/croissant/kinds/billboard.yaml
JackPlaymint: Screenshot_2024-02-23_at_13 19 44

map/Assets/Scripts/Components/Billboard/BillboardData.cs
```diff
@@ -0,0 +1,10 @@
+public class BillboardData : BaseComponentData
+{
+    public int q;
+    public int r;
+    public int s;
+    public float height;
+    public float rotation;
+    public string? selected; // none/highlight/outline
+    public string? url;
```
chrisfarms, line 9: `url` is not very descriptive ...and potentially inacurate if we support data uris ... maybe `imageUri` or even just `image` would be better to hint that it's not just a link or something

cli/src/utils/manifest.ts
```diff
@@ -124,6 +124,14 @@ export const BuildingKindExtractorSpec = z.object({
     outputs: Slot.array().max(1).optional(),
 });
 
+export const BillboardSpec = z.object({
+    category: z.literal('billboard'),
+    name: Name,
+    description: OneLiner.optional(),
+    model: z.string(),
```
chrisfarms, line 131: can we restrict this model field to just predefined values that represent potential models so that it is consistent with the others and gives decent validation errors 🙏 

something like:

```
const BillboardModel = z.enum(['billboard', 'computer']);
...
model: BillboardModel
```

contracts/src/maps/croissant/kinds/billboard.yaml
chrisfarms, line 7: as per other comment ... this field should be from a fixed set ie "billboard"

frontend/src/components/map/Buildings.tsx
```diff
@@ -133,6 +134,19 @@ export const Buildings = memo(
...
                             />
                         );
+                    } else if (getBuildingCategory(b.kind) == BuildingCategory.BILLBOARD) {
+                        return (
+                            <Billboard
+                                key={b.id}
+                                id={b.id}
+                                height={height}
+                                rotation={rotation}
+                                url={b.kind?.model?.value}
```
chrisfarms, line 144: heh ok I see you are using the model just as a way to set the url quickly.

we will probably need this to get passed in via a pluginMapProperties type thing just like we do for tile colors:

https://github.com/playmint/ds/blob/main/frontend/src/components/map/Tile.tsx#L71

map/Assets/Scripts/Components/Billboard/BillboardData.cs
JackPlaymint: I've switched it to "image" for now. I checked passing a data uri and it worked! So that's once less thing to worry about :)

5p0rt5BEArD (commented):
Latest code changes look good and address the comments - just going to test it locally before approving

5p0rt5BEArD (approved):
All works in local test.

I noticed that placing more than one billboard results in only one displaying the cocktail. Could make the cocktail hut set the image on all billboards but that might be just as weird - choice of billboard is left as an exercise for the reader

Issue #1067, opened 2024-02-20
allow seeing ERC1155-items in metamask UI


chrisfarms:
there is an API to tell metamask about tokens to display ... we should use it to allow people to see their wallet-owned downstream-items

Issue #1056, opened 2024-02-16
erc1155 compatible item exchange


5p0rt5BEArD:
we want to make it possible to transfer items from unit inventories into a player wallet inventory ... items in this wallet inventory act like standard erc1155 tokens

Issue #1062, opened 2024-02-20
prompt user to add/switch networks if provider detected on wrong network


chrisfarms:
we want to ensure any connected wallet providers have the correct network selected

so that we can start doing some things that interact with the network directly (like minting tokens that can be seen in wallets)

for kick off:

* each hexwood0....hexwood16 deployment has it's own RPC network endpoint
* do we need to give them each unique chain ids too?
* this doesn't apply to the throwaway/burner providers
* focus on metamask, but maybe investigate how walletconnect providers will be affected, potentially spinning out future task

PR #1081, opened 2024-02-26
import / export items as ERC1155 tokens

chrisfarms:
### what

* Allow players to import and export downstream items into an ERC1155 compatible form for use outside of the main game world
* Players get a kind of "global inventory" which they can move items into or out of, when items are in this inventory they behave as standard ERC1155 tokens, and can be transferred.

https://github.com/playmint/ds/assets/45921/856930fa-3e35-45fa-ab39-df08a2b6606f

### why

* we believe having a mechanism for exchanging tokens outside of the main game will give additional surface area for composability/interoperability

### how

There are two new game actions:

* `EXPORT_ITEM`  - transfer an item you have control over in game to an ethereum address
* `IMPORT_ITEM` - transfer an item you have in your player wallet into an in-game inventory

### testing with metamask

👀 **Please read!** ... metamask integration is flakey

You can test this out and play with the metamask bits, but you must:

* Set "display nft media" to `true` ... in metamask `settings -> security/privacy` 
* You must manually add the localhost network  and ensure you are using it (until #1062 is done)
* You must clear all the activity/nonce history each time you start `make dev`
* Metamask lags behind the network alot ... you will need to spam the "add token to metamask" button alot before it eventully works
* Any tokens you add may not update their state until you click on them... so they can get out of sync easily
* **this one is the best** ... there's a bug in the "send" feature, it will ask you to type in a balance to send ... it only lets you type in decimal, but behind the scenes it sends as hex ... so sending `10` will send `16` .... tbh they do say they don't actually support sending, so guess it's a work in progress ... but probably stick to sending small numbers to avoid confusion
* the way metamask queries the chain via old state means I had to disable anvil's prune-history 😱  .... this is BAD as prune-history is the thing that is keeping anvil's disk/memory usage low
* you will also need some ETH to do a transfer ... give yourself some eth with `cast send --private-key 0x8451fe83595a88521cac265c1356568ab84c2b5d81654f36ea309be28ded5a0d --value 5.0ether $YOUR_ADDRESS` (replacing the YOUR_ADDRESS bit)

### related

* resolves: #1056 
* resolves: #1067 

### todo

before merge...

* [x] resolve issue where pending state does update on transfer
* [x] what to do about prune-history
* [x] check wallet-inventory works alongside quests panel
* [x] tidy up the text in the wallet inventory panel
* [x] stop throwing in inventory failure

### DIFF
--- a/contracts/src/Items1155.sol
+++ b/contracts/src/Items1155.sol
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
 pragma solidity ^0.8.13;
 
-import "forge-std/console2.sol";
 import "cog/IState.sol";
 import {Schema, Node, Kind, GOO_GREEN, GOO_BLUE, GOO_RED} from "@ds/schema/Schema.sol";
 import "@ds/utils/Base64.sol";
@@ -97,7 +96,6 @@ contract Items1155 is ERC1155 {
         bytes calldata data
     ) public virtual override {
         super.safeBatchTransferFrom(from, to, ids, amounts, data);
-        console2.log("------safeBatchTransferFrom");
 
         for (uint256 i = 0; i < ids.length; i++) {
             _updateOwnedToken(from, to, ids[i], amounts[i]);
@@ -106,11 +104,6 @@ contract Items1155 is ERC1155 {
 
     function _updateOwnedToken(address from, address to, uint256 tokenId, uint256 amount) public virtual {
         bytes24 itemId = bytes24(uint192(tokenId));
-
-        console2.log("to", to);
-        console2.log("tokenId", tokenId);
-        console2.log("addr", address(this));
-        console2.log("amount", amount);
 
         // decement from
         if (from != address(0)) {
--- a/contracts/src/Items1155.sol
+++ b/contracts/src/Items1155.sol
@@ -105,7 +105,7 @@ contract Items1155 is ERC1155 {
     function _updateOwnedToken(address from, address to, uint256 tokenId, uint256 amount) public virtual {
         bytes24 itemId = bytes24(uint192(tokenId));
 
-        // decement from
+        // decrement from
         if (from != address(0)) {
             bytes24 ownedTokenId = Node.OwnedToken(tokenId, Node.Player(from));
             (bytes24 existingItem, uint64 existingBalance) = state.getItemSlot(ownedTokenId, 0);
--- a/contracts/src/Items1155.sol
+++ b/contracts/src/Items1155.sol
@@ -1,0 +1,138 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.13;
+
+import "forge-std/console2.sol";
+import "cog/IState.sol";
+import {Schema, Node, Kind, GOO_GREEN, GOO_BLUE, GOO_RED} from "@ds/schema/Schema.sol";
+import "@ds/utils/Base64.sol";
+import "@ds/utils/LibString.sol";
+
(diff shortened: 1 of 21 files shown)


contracts/src/Items1155.sol
```diff
@@ -0,0 +1,131 @@
...
+
+        for (uint256 i = 0; i < ids.length; i++) {
+            _updateOwnedToken(from, to, ids[i], amounts[i]);
+        }
+    }
+
+    function _updateOwnedToken(address from, address to, uint256 tokenId, uint256 amount) public virtual {
+        bytes24 itemId = bytes24(uint192(tokenId));
+
+        // decement from
```
5p0rt5BEArD, line 108:
```suggestion
        // decrement from
```

contracts/test/rules/InventoryRule.t.sol
```diff
@@ -47,23 +47,23 @@ contract InventoryRuleTest is Test, GameTest {
             mobileUnit, // location of from-bag
             mobileUnit // location to to-bag
         );
-        vm.stopPrank();
+        // vm.stopPrank();
```
5p0rt5BEArD, line 50: Why are all these commented out?

5p0rt5BEArD (commented):
Should we have some test code for the new import / export actions?

contracts/test/rules/InventoryRule.t.sol
chrisfarms: I was testing with a newer foundry/forge ... where they have changed how stopPrank works, and it's now an error to call it twice (which all these things do due to the implied stop at the end).

I will drop this commit, and go back to our pinned version of forge

chrisfarms:
> Should we have some test code for the new import / export actions?

yeah we should ... will add a couple

contracts/test/rules/InventoryRule.t.sol
chrisfarms: _dropped_

5p0rt5BEArD (commented):
Code looks good.
Works locally for me up to the point of sending the NFT - my metamask doesn't let me.

So going to try and work out how to send to another wallet so I can test the flow of items coming into the game

Screenshot at Feb 26 14-45-48

chrisfarms:
> Works locally for me up to the point of sending the NFT - my metamask doesn't let me.

Oh I wonder if I'm in some kind of A-B test or canary rollout then ... that would explain why I can send event tho the docs say I shouldn't be able to ... and also why it's so broken heh.

Well you could do it manually with `cast` if you wanted to check.

* first add the private key `0x8451fe83595a88521cac265c1356568ab84c2b5d81654f36ea309be28ded5a0d` to your metamask
* signin to downstream using that account
* export 100 green goo into your wallet-inventory
* now you can do a transfer directly by crafting a transfer tx using `cast`:

```
cast send --private-key 0x8451fe83595a88521cac265c1356568ab84c2b5d81654f36ea309be28ded5a0d 0x0a668c3342e68b5F9054403e2080fA77259e8DbA 'safeTransferFrom(address, address, uint256, uint256, bytes)' 0x4208a6518500E980ED44Da94ea31b85c85ec4568 $TARGET_ADDRESS 2610836604245033138765654868554422916039452987239149076480 1 ""
```

replace $TARGET_ADDRESS with address of where to send 1 green goo

5p0rt5BEArD:
> Oh I wonder if I'm in some kind of A-B test or canary rollout then ... 

Updating metamask fixed it for me.
I had to enable developer mode in my browser extension manager to get access to the Update button.

5p0rt5BEArD (approved):
tested and working localy

PR #1085, opened 2024-02-26
per-deployment chain-ids + auto switch metamask to correct chain id

chrisfarms:
### what

- include network id,endpoint,name in config.json
- dynamically generate unique chain-id based on namespace for each deployed network
- automatically add/switch metamask to the configured endpoint when you select "metamask"

Screenshot 2024-02-27 at 09 08 50

### why

to make it easier to play with things like Items as ERC1155 tokens where you would need to connect to the correct network endpoint to interact

we cannot add different network endpoints to metamask without also using different chain ids ... so need to spin up the networks with different chain id and then pass this configuration into the app so that it can tell metamask the correct endpoint + chain id

### related

* resolves: #1062 

### issues

* if the switch fails, we just mark the attempt as failed and never try again ... otherwise we can get stuck in a loop of attempting.... it might be better to present a "switch network" button, but that will mean changing the UI a fair bit more
* it will only ever prompt to switch when you click "metamask" in the connect dialog ... it doesn't detect if you have manually switched networks after than point ... so it's still easy for it to get out of sync
* we could/should support switching walletconnect wallets too, but the focus here has been on metamask and demoing ERC1155 tokens in metamask so this is out of scope for now

### before merge

check that it correctly adds/switches when running...

* [x] via make dev
* [x] via compose up
* [x] via a deployed hexwoodN instance (switching works but because history is pruned metamask things like ERC1155 tokens don't work)

### when testing...

the game tries to cache your account so that you don't have to sign in all the time ... this means you will likely need to "disconnect" and refresh before you get prompted to add or switch network

### DIFF
--- a/.devstartup.js
+++ b/.devstartup.js
@@ -42,7 +42,7 @@ async function handleStartup(){
     const commands = [
         {
             name: 'networks',
-            command: "anvil -m 'thunder road vendor cradle rigid subway isolate ridge feel illegal whale lens' --block-time 2 --block-base-fee-per-gas 1 --transaction-block-keeper 250",
+            command: "anvil --chain-id 22300 -m 'thunder road vendor cradle rigid subway isolate ridge feel illegal whale lens' --block-time 2 --block-base-fee-per-gas 1 --transaction-block-keeper 250",
             prefixColor: 'black',
         },
 
--- a/chart/templates/app.yaml
+++ b/chart/templates/app.yaml
@@ -3,6 +3,9 @@ gameID: {{ $.Values.frontend.gameAddress | quote}}
 build: {{ $.Values.version | quote }}
 wsEndpoint: "wss://services-{{ $.Release.Namespace }}.{{ $.Values.cluster.domain }}/query"
 httpEndpoint: "https://services-{{ $.Release.Namespace }}.{{ $.Values.cluster.domain }}/query"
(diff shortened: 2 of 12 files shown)


.devstartup.js
```diff
@@ -42,7 +42,7 @@ async function handleStartup(){
     const commands = [
         {
             name: 'networks',
-            command: "anvil -m 'thunder road vendor cradle rigid subway isolate ridge feel illegal whale lens' --block-time 2 --block-base-fee-per-gas 1 --transaction-block-keeper 250",
+            command: "anvil --chain-id 22300 -m 'thunder road vendor cradle rigid subway isolate ridge feel illegal whale lens' --block-time 2 --block-base-fee-per-gas 1 --transaction-block-keeper 250",
```
5p0rt5BEArD, line 45: How do you choose the chain number?
Do you check https://chainid.network/?

5p0rt5BEArD (approved):
Code all looks very neat
Tested locally and on the hexwood1 deploy and metamask successfully added network (chainids 22300 and 22301,named hexwoodlocal and hexwood1) and switched to the network
Limitations explained in pr description make sense and I agree are not priorities to try and improve

gold star ⭐

.devstartup.js
chrisfarms: > How do you choose the chain number?

arbitrarily .... 😅 

....but yeah I checked to make sure nothing popular is out there yet using the 22300 - 22316 range

Issue #1087, opened 2024-02-27
plugins slow and stalling


chrisfarms:
reports of plugins stopping updating things:

* counters not updating in DvB
* tonk action buttons not appearing
* some building plugins taking a very long time to arrive

these could all be symptoms of plugins crashing or not responding ... this is NOT the same as the VM out of memory, we see no OOM errors, things just stall and require a refresh

potential clue:

* this might be related to the size of the state ... we are spending ALOT of time in state serialization again 👀

Issue #1086, opened 2024-02-27
tonk UI shows even when you don't have a tonk


chrisfarms:
... I think the code that decides when to load item plugins only looks at the slot item id and NOT the balance ... so a balance=0 still counts

PR #1089, opened 2024-02-27
Snowsphere: the unsafe sandbox

chrisfarms:
### what

This PR is attempting to resolves issues with flakey and unresponse plugin execution

...the main part is:

* cleans up the interface between the game and the sandbox
* adds (and enables) a new totally unsafe implementation of the sandbox interface

...and then also some bug fixes to improve the general plugin situation:

* ensure that plugins never execute concurrently ... it was unexpected that they were doing this
* add a max execution time timeout so that slow plugins don't cause an ever growing backlog (currently 1s)
* fix issue where item plugin would be loaded even if balance of the item is 0
* try to stop tonk plugin crash looping by catching errors in the plugin and fixup a couple of bugs
* stop storing old session data ... there's a chance this is causing the "cant craft" issue
* stop using random plugin function ref ids .... there's a chance this is causing the "cant craft" issue

### why

we are having problems with plugins executing within the sandbox. 

it is very likely these issues are due to the large memory requirements of serializing and process the large `state` object that is passed into plugins over and over again, but debugging and visibility is poor within the sandbox vm. this un-sandbox let's us execute plugins "normally" and lets us use the profiling tooling directly.

it also means plugin loading and state serialization is drastically reduced, so could be used as a stop-gap until we decide to fix the state serialization problem properly

### related

* fixes: #1086 
* fixes: #1087 

### before merge (if we even want that!)

* [x] test with lots of plugins 
* [x] have a team playtest with this branch on croissant map
* [x] support the new ds exports in https://github.com/playmint/ds/pull/1097

### DIFF
--- a/frontend/src/workers/snowsphere.ts
+++ b/frontend/src/workers/snowsphere.ts
@@ -1,0 +1,243 @@
+// ///////////////////////////////////////
+// snowsphere is the opposite of sandbox
+//
+// it is completely insecure!
+//
+// plugins executed with snowsphere has unrestricted
+// access to basically everything.
+//
+// they can read your indexdb
+// they can look at your cookies
+// they can start workers
+// they can request stuff as if they are the origin
+// they can probably steal the session key
+// they can probably get access to the DOM if they try hard enough
+// they can probably turn up at your house
+//
+// basically this offers no protection whatsoever
+//
+// but ... they are fast
+// and they can use the normal debugging instrumentation
+// so that's nice
+//
+// ////////////////////////////////////////
+//
+import {
+    GameConfig,
+    GameStatePlugin,
+    Logger,
+    PluginConfig,
+    PluginDispatchFunc,
+    PluginUpdateResponse,
+    Sandbox,
+    actionArgFromUnknown,
+} from '@downstream/core';
+import * as Comlink from 'comlink';
+import { ethers } from 'ethers';
+
+let globalPluginState: GameStatePlugin;
+let globalPluginBlock: number = 0;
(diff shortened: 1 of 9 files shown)


frontend/src/workers/snowsphere.ts
```diff
@@ -0,0 +1,225 @@
...
+    logMessage: Logger,
+    questMessage: Logger,
+    config: PluginConfig
+): Promise<number> {
+    const id = ++globalContextSeq;
+    const code = rewriteModuleCode(config.src);
+    const mod = await import(/*webpackIgnore: true*/ esm(code));
+
+    // all the stuff available via ds.etc
+    mod.__setDS({
```
chrisfarms, line 146: @hypnoshock has raised https://github.com/playmint/ds/pull/1097 to expose some more functions on the ds module.

so I'll need to update this in snowsphere after that PR is merged

frontend/src/workers/snowsphere.ts
```diff
@@ -0,0 +1,247 @@
...
+// plugins executed with snowsphere has unrestricted
+// access to basically everything.
+//
+// they can read your indexdb
+// they can look at your cookies
+// they can start workers
+// they can request stuff as if they are the origin
+// they can probably steal the session key
+// they can probably get access to the DOM if they try hard enough
+// they can probably turn up at your house
```
hypnoshock, line 15: "No plz, I have a wife and children!"

hypnoshock (approved):
It's so much simpler and easy to read than the version passing args around as json.
I approve the door wide open plugin system for the moment.

PR #1093, opened 2024-02-28
Door building and model swapping

JackPlaymint:
This PR adds a building that represents a door and allows the ability to swap a building's model using a plugin.

Jimmy has made an example of the Door's functionality in contracts/src/maps/example-gate.
In the example there's a bag with a key in it and a locked door which you cannot move past.
When you place the key in your inventory, the door will open and you can navigate through.

The Gate.js plugin code handles swapping the door model and changing the tile colour and type (to make it a blocker or not).

The door includes a new 3D asset from Tom which I have set up as two separate building models (open and closed).

When opening the closed model type, a brief animation is played before the actual model is swapped, for all other models the swap happens instantly.

There is some work to be done to allow the user to specify a 'left' or 'right' rotation for the door, however it has been decided that all buildings should have this restriction/functionality so I'll add it in as a further task.


frontend/src/components/map/Buildings.tsx
```diff
@@ -76,6 +76,10 @@ export const Buildings = memo(
                     const selected = selectedElementID === b.id ? 'outline' : 'none';
                     const rotation = lerp(-20, 20, 0.5 - getUnscaledNoiseFromCoords(coords));
                     const tile = tiles.find(({ id }) => id === b.location?.tile.id);
+                    const overrideModel = randomTileProperties
```
chrisfarms, line 79: **non-blocking**: we should rename `randomTileProperties` to something that makes more sense like `pluginBuildingProperties`

map/Assets/Scripts/Components/FactoryBuilding/FactoryBuildingController.cs
```diff
@@ -64,12 +70,25 @@ protected void Update()
...
             {
-                foreach (FactoryBuildingBlockController block in _blocks)
+                if (_nextData.model.Contains("door"))
+                {
+                    GameObject prefab = overridePrefabs.FirstOrDefault(
+                        (p) => p.name == _nextData.model
+                    );
+                    if (_currentAnimator != null && _nextData.model.Contains("open"))
+                    {
+                        StartCoroutine(WaitForAnimationEndCR(prefab, "DoorUnlock"));
```
chrisfarms, line 80: If I'm switching from a `01-01` to `door-open` ... then does this animation essentially just delay they swap? ... I don't think that's a problem, just trying to play it in my head

chrisfarms (approved):
this is fun :) 

spun it up ... works as advertised.

map/Assets/Scripts/Components/FactoryBuilding/FactoryBuildingController.cs
JackPlaymint: It should be ok, as model 01-01 doesn't have an animator, it should ignore the animation delay coroutine and go straight to the SwapModels function.

The door opening animation is actually applied to the door-closed model, and it's currently the only model with an animator, so the delay should only happen in the correct circumstance.

frontend/src/components/map/Buildings.tsx
JackPlaymint: Good thinking, pushed that change up now

PR #1106, opened 2024-03-01
Added `offset` command to cli

hypnoshock:
# What

- Added `offset` command to cli to offset location fields in manifests by a given coordinate. 
- Result is output to stdout
- Works with `-R` flag

example:
`ds offset 15,0,-15 -R -f contracts/src/maps/example-gate/ > ~/temp/offset.yaml `

Below I've deployed the example-gate map with different offsets to get two copies of it in the world

image

# Why

Allows us to compose a large map from smaller maps without having to manually edit each location field

# Note

I had a problem with coords that started with a negative number as it tripped up the command line parser. To get round this you can specify QRS coords like this: 
`ds offset "[-10,10,0]" -R -f contracts/src/maps/example-gate/ > ~/temp/offset.yaml`

### DIFF
--- a/cli/src/commands/apply.ts
+++ b/cli/src/commands/apply.ts
@@ -3,10 +3,7 @@ import { globSync } from 'glob';
 import path from 'path';
 import { z } from 'zod';
 import { Op, getOpsForManifests } from '../utils/applier';
-import {
-    ContractSource,
-    readManifestsDocumentsSync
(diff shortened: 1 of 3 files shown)


chrisfarms (approved):
handy 

wonder if it shouldn't be a top level command, as it's kind niche ... but don't have a good suggestion

Issue #1103, opened 2024-03-01
tidy up the croissant map into a croissant shape


chrisfarms:
... make the croissant map more pleasing to look at .... shift hexcraft up a bit etc

PR #1107, opened 2024-03-01
Croissant map

PlaymintPaul:
Putting the croissant into the map. Moved the games around a bit

resolves: #1103 
resolves: #1100


5p0rt5BEArD (approved):
Tested a few times and working with all games
Billboards need placing but that's not a blocker

PR #1108, opened 2024-03-01
tile fabricator, shift tiles with WASDQE keys

5p0rt5BEArD:
### What
The entire map can be moved via key presses in the tile-fabricator

### Why
We have some complex map shapes that we sometimes want to create a new map with but offset. This should help with that.

### How
Mapped WASDQE keys to:
A = left
D = RIght
W = Up (right)
S = Down (left)
Q = Up (left)
E = Down (right)


chrisfarms (approved):
wen vim keybindings

PR #1109, opened 2024-03-01
Optimised Combat Finalisation

### LINKED ISSUE #1095: issues with large number of updates getting queued
10+ players on DvB got stuck when everyone attempting to build/finalize-combat at the same time

network requests were getting cancelled, optimistic updates got out of sync .... until it eventually all wave colapsed back to reality after like 90s or so

Steedie:
## What
This change intends to make it so that the last unit that moved is **only** & most likely to be the one to send the `FINALISE_COMBAT` action.

In this screenshot, combat vs a building has just finished, and unit#193d is the only one who attempted to send the action because this was the last unit that moved:
image

## Why
While playtesting Duck Burger with 10+ players, move of our clients were not updating throughout the game, and we think this was caused by everyone trying to execute the `FINALISE_COMBAT` action at the same time. (See [this comment](https://github.com/playmint/ds/issues/1095#issue-2160984833))

## How
It ranks all units on the map by when they last moved, placing the unit that moved closest to the current world block time first:
```js
const sortedMobileUnits =
            world?.mobileUnits?.sort((a, b) => {
                const diffA = Math.abs((a.nextLocation?.time || 0) - blockNumber);
                const diffB = Math.abs((b.nextLocation?.time || 0) - blockNumber);
                return diffA - diffB;
            }) || [];
```

Using this, it calculates a sleep time, and uses setTimeout to make sure the most recently active units are prioritized and sequentially try to send the action:
```js
const jitter = Math.random() * 1000; // For cases where multiple units moved on the same block
        const sleepFor = selectedIndex > -1 ? selectedIndex * (2000 + jitter) : 0;
```

Therefore, if another client is running this `useEffect` but has more of a delay than another client, it won't try to send the action because the combat at that point is already finalized.

Resolves #1095

### DIFF
--- a/frontend/src/components/views/shell/index.tsx
+++ b/frontend/src/components/views/shell/index.tsx
@@ -328,6 +328,7 @@ export const Shell: FunctionComponent<ShellProps> = () => {
                     }
                     return () => clearTimeout(timeoutId);
                 }
+                return;
             }, sleepFor);
         }
     }, [combatSessionTick, player, prevCombatSessionTick, blockNumber, selectedMobileUnit, world?.mobileUnits]);
--- a/frontend/src/components/views/shell/index.tsx
+++ b/frontend/src/components/views/shell/index.tsx
@@ -79,17 +79,14 @@ export const Shell: FunctionComponent<ShellProps> = () => {
     }, [player?.quests]);
     const unfinalisedCombatSessions = useMemo(
         () =>
-            (world?.sessions || [])
-                .filter((s) => {
-                    return !s.isFinalised;
-                })
-                .filter((s) => {
-                    const oneSideZero =
-                        s.attackers.filter((paticipant) => paticipant.node.id != nullBytes24).length == 0 ||
-                        s.defenders.filter((paticipant) => paticipant.node.id != nullBytes24).length == 0;
-                    const combatStarted = blockNumber && blockNumber >= (s.attackTile?.startBlock || 0);
-                    return oneSideZero || combatStarted;
-                }),
+            (world?.sessions || []).filter((s) => {
+                const isNotFinalised = !s.isFinalised;
+                const oneSideZero =
+                    s.attackers.filter((paticipant) => paticipant.node.id != nullBytes24).length == 0 ||
+                    s.defenders.filter((paticipant) => paticipant.node.id != nullBytes24).length == 0;
+                const combatStarted = blockNumber && blockNumber >= (s.attackTile?.startBlock || 0);
+                return isNotFinalised && (oneSideZero || combatStarted);
+            }),
 
         [world?.sessions, blockNumber]
     );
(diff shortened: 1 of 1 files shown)


5p0rt5BEArD (commented):
Just tried a local test and it took about 20 seconds for combat to visually finalise
I need to go and compare with the behaviour before this change

Steedie:
> Just tried a local test and it took about 20 seconds for combat to visually finalise I need to go and compare with the behaviour before this change

How many units did you have in the map when you tried this?

I just tried with 2 units and closed the tab of the most recently moved one and it didn't finalise until I moved the other unit (~30 seconds after combat should have finished) so I'll look into these issues now.

Steedie:
> Just tried a local test and it took about 20 seconds for combat to visually finalise

My latest commit should now fix this problem. (Changed back to _ready for review_)

frontend/src/components/views/shell/index.tsx
```diff
@@ -274,24 +274,64 @@ export const Shell: FunctionComponent<ShellProps> = () => {
...
-            Promise.all(finaliseActions.map((action) => player.dispatchAndWait(action))).catch((err) =>
-                console.warn(err)
-            );
+                        // Dispatching all at once could prevent future combat sessions from finalising if
+                        // one of the finalise transactions fail so we dispatch them one by one
+                        Promise.all(finaliseActions.map((action) => player.dispatchAndWait(action))).catch((err) =>
+                            console.warn(err)
+                        );
+                    }
+                    return () => clearTimeout(timeoutId);
```
hypnoshock, line 329: I might be reading this wrong, but I don't think returning the `clearTimeout` here is not going to do anything as it's returning a function inside the timeout callback. Are we trying to clear the timeout if someone else manages to finalise the combat?

hypnoshock (approved):
Looks alright so approved but I'm a bit unsure of what the `clearTimeout` is going to do as it's within the scope of the `setTimeout` callback. I think the idea is to clear the timeout if someone else has finalised the combat but I think as it stands at the moment, everyone will still send a finalise request just staggered over time.

frontend/src/components/views/shell/index.tsx
Steedie: > I think the idea is to clear the timeout if someone else has finalised the combat

It was originally there to solve a problem where it would use the same `finaliseActions` list multiple times, and it seemed to fix that issue at the time. I've since changed the ordering quite a few times so it may not be doing what I originally intended it to do.

> everyone will still send a finalise request just staggered over time

This should only be the case for up to 3 units since it should be getting the latest state of the world when it attempts to finalise combat, and should only send the action if there is anything that _still_ needs to be finalised. (After the 3rd most active unit, the sleep time is enough that the new world state is ready)

I'll try taking out the clearTimeout to see if it still behaves the same way, and I'll make the console.log more clear since it's possible for more than 1 units to try and send the `FINALISE_COMBAT` action.

PR #1111, opened 2024-03-04
Duck Burger: 3 Minute Timer by Default

### LINKED ISSUE #1088: Duck Burger Timer should default to 3 minutes

Steedie:
## What
Default Duck Burger game time set to 3 minutes instead of 1.

## Why
So that the Duck Burger play tests can be a little bit less chaotic.

## How
```solidity
// set endblock to now plus 3 minutes (assuming 2 second blocks)
_setDataOnBuilding(dispatcher, buildingId, "endBlock", bytes32(block.number + 3 * 30));
```

Resolves #1088


5p0rt5BEArD (approved):
yep :)

PR #1116, opened 2024-03-05
Updated building fabricators 'BasicFactory' bytecode

### LINKED ISSUE #1113: I can't place a building created in the Fabricator
Trying to place a building I just created in the Fabricator on Hexwood6. I can see the building on the list and I have enough resources, but when I click on confirm nothing shows up.

https://github.com/playmint/ds/assets/129759334/8b9c7939-4c83-4708-9111-af4a34df4e30

hypnoshock:
# What

The bytecode that the building fabricator was deploying wasn't the correct bytecode for the hardcoded contract code

# Why

Incorrect code would deploy but when contructing the deployed building on the map, the contract would revert

fixes #1113


chrisfarms (approved):
I have reviewed the bytes one-by-one ... I can confirm they are bytes

I assume this is due to the change for "gate buildings" which changed the base contract.

do we need any updates to the hardcoded JS too? it looks like it compiles fine, so probably fine.

hypnoshock:
> I have reviewed the bytes one-by-one ... I can confirm they are bytes
> 
> I assume this is due to the change for "gate buildings" which changed the base contract.
> 
> do we need any updates to the hardcoded JS too? it looks like it compiles fine, so probably fine.

There isn't anything to update with the JS and yes I think you're right that it was introduced when the base contract changed. You can see that the base contract bytecode is about 5 times the size from the previous version. I'm not sure when we last checked the building fabricator but it might even be that the bytecode wasn't updated when we added the on construct hook either!

PR #1119, opened 2024-03-06
Croissant Map Iterations - Multiple changes

PlaymintPaul:


5p0rt5BEArD:
Addresses issue #1120

5p0rt5BEArD:
Some points that I could just do myself if you agree with them:

- I included adding a building to make use of id card and letter in the issue
- This could be the cupis
- The apex needs moving ;)

image


image

PlaymintPaul:
>     * The apex needs moving ;)

Eh?! Mine is much closer to the image than your proposed one :)

5p0rt5BEArD (approved):
We played as a team and all looks good

PR #1124, opened 2024-03-06
README: Note for Windows users to enabled symlinks

Steedie:
## What
Updated README to include a note for Windows users to enable symlinks

## Why
Paul and I had an error on our laptops where it was not recognizing symlink files.

## How
To solve this problem, run `git config core.symlinks true` in `ds` dir.


5p0rt5BEArD (approved):
Nicely worked out

PR #1129, opened 2024-03-06
Unit model swapping

JackPlaymint:
This PR adds the ability to change a Unit's model from a plugin.

There are 15 different models to choose from, although only a few with hoodie colour variations are currently used.
The IDs for the 15 models are:

- Unit_Astronaught_01
- Unit_Astronaught_02
- Unit_Chicken_01
- Unit_Chicken_02
- Unit_Engineer_01
- Unit_Engineer_02
- Unit_Hoodie_01
- Unit_Hoodie_02
- Unit_Hoodie_03
- Unit_Hoodie_04
- Unit_Hoodie_05
- Unit_Hoodie_06
- Unit_Hoodie_07
- Unit_Tuxedo_01
- Unit_Tuxedo_02

There are some examples in the croissant games of how to change the models. The models change at various points in the games instead of directly changing the material colour as we used to.
I've kept the parameters for sending colour from the frontend to the Unity models, but the data is currently unused, there's potential to use it in the future to change accent colours on certain models if supported.


5p0rt5BEArD (approved):
Code looks good and works in local tests

The new detail models are affected by the thick highlight outline more than the old grey ones. Not something to fix iin this pr though

Issue #1135, opened 2024-03-07
model switching sometimes causes unity crash


chrisfarms:
.... probably a null check missing

PR #1136, opened 2024-03-07
Fix for null model bug

JackPlaymint:
This PR fixes a null reference error which could occur when swapping the unit models sometimes.
The logic for handling resetting a model back to default was flawed, but is now more inline with how other model/color swaps work.

resolves: #1135

### DIFF
--- a/map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
+++ b/map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
@@ -49,19 +49,11 @@ public class MobileUnitController : BaseComponentController<MobileUnitData>
 
         if (_nextData.model != null)
         {
-            if (
-                _prevData == null
-                || _nextData.model != _prevData.model
-                || _nextData.color != _prevData.color
-            )
+            if (_nextData.model != (_prevData?.model ?? "Unit_Hoodie_07"))
             {
                 DestoryPreviousModels();
                 ShowUnitModel(_nextData.model);
             }
-        }
-        else if (_prevData.model != null && _nextData.model != _prevData.model)
-        {
-            ShowUnitModel("Unit_Hoodie_07"); //Reset model
         }
 
         // movement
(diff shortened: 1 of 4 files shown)


map/Assets/Scripts/Components/MobileUnit/MobileUnitController.cs
```diff
@@ -49,20 +49,12 @@ protected void Update()
 
         if (_nextData.model != null)
         {
-            if (
-                _prevData == null
-                || _nextData.model != _prevData.model
-                || _nextData.color != _prevData.color
-            )
+            if (_nextData.model != (_prevData?.model ?? "Unit_Hoodie_07"))
```
chrisfarms, line 52: Not sure I understand the `?? Unit_Hoodie_07` bit here?

Is this an artifact of the ShowUnitModel being called in Start?

Maybe I'm misunderstanding something but ... would be better to avoid calling ShowUnitModel in `Start`, and then just let this update call it with the correct model on first update, to avoid loading the wrong model being loaded for a split second?

JackPlaymint: Ah yeah, I see what you mean. My thinking was to always show a model, but you're right it's probably better to show no model for a second or two rather than the wrong one.

map/Assets/Scripts/Components/MobileUnit/MobileUnitModelController.cs
```diff
@@ -33,6 +51,8 @@ public class MobileUnitModelController : MonoBehaviour
             t += Time.deltaTime * deltaMultiplier;
             foreach (Renderer rend in renderers)
             {
+                if (rend == null) // This shouldn't happen, but I'm scared now.
```
chrisfarms, line 54: 😆

map/ProjectSettings/ProjectSettings.asset
```diff
@@ -764,7 +764,7 @@ PlayerSettings:
   blurSplashScreenBackground: 1
   spritePackerPolicy: 
   webGLMemorySize: 32
-  webGLExceptionSupport: 0
+  webGLExceptionSupport: 3
```
chrisfarms, line 767: Can you disable this 🙏 ... it's a big performance hit

chrisfarms (changes requested):

map/ProjectSettings/ProjectSettings.asset
JackPlaymint: Whoops! Yep

chrisfarms (approved):

PR #1143, opened 2024-03-08
Rockaway bug fixes

hypnoshock:
# What

- Fixed resetting of the map by relatively positioning buildings / bags from the Labyrinth Core building
- Fixed problem where the croissant version of Labyrinth was using the common version of Large Rocks. The common version has lower stats which makes the game uncompletable so I have added the Large rocks to their custom.yaml and renamed both very large and large rocks to 'boulders' so they don't clash.
- Removed all the unused functions from LabyrinthCore just as a tidy up

### DIFF
--- a/contracts/src/maps/croissant/maps/labyrinth/room_4.yaml
+++ b/contracts/src/maps/croissant/maps/labyrinth/room_4.yaml
@@ -1,7 +1,7 @@
 ---
 kind: Building
 spec:
-  name: Very Large Rocks
+  name: Very Large Boulders
   location:
     - 21
     - -27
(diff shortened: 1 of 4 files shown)


5p0rt5BEArD (approved):
Tested and it all works
Now I see that there is just one bag spawned on the reset it may have been working before anyway - sorry!
I was expecting all the labyrinth bags to be replaced but its just putting the crucial items back.

I thin there's a risk of it becoming hard to play (or needing to resort to the wallet) if we don't have all those bags.
Also I think we need a hidden reset option in case some stops half way through and doesn't reset. These are new issues that may not be worth doing right now anyway. So all good

PR #1145, opened 2024-03-08
Key check bug fix for labyrinth

### LINKED ISSUE #1123: Stuck in room 2 of Labyrinth
With both the first two keys in my Unit inventory I could do not move back through the second door
Frontend was happy I could but contract reverted 

https://github.com/playmint/ds/assets/90266137/3f8cf6d6-5c77-4676-93ad-21774240a4e8

5p0rt5BEArD:
### What
Moving a key from unit inventory to wallet inventory and back could break the labyrinth key check code if using a different slot when moving it back.

This is fixed.

### How
In the gate solidity code, we now confirm an item in a slot actually has a balance before declaring it found.

This is becessary because there is a bug where an item removed from a slot can leave a 0 balance item entry for that item.

fixes #1123 

before:
https://github.com/playmint/ds/assets/90266137/7eb13c23-a97d-4b47-addd-b410d7c736ec

after:
https://github.com/playmint/ds/assets/90266137/229b7e36-cf81-44e9-a652-04eeecce06a1

### DIFF
--- a/contracts/src/maps/labyrinth/kinds/buildings/Gates/Gate.sol
+++ b/contracts/src/maps/labyrinth/kinds/buildings/Gates/Gate.sol
@@ -51,7 +51,7 @@ abstract contract Gate is BuildingKind {
             for (itemSlot = 0; itemSlot < 4; itemSlot++) {
                 bytes24 bagItemId;
                 (bagItemId, balance) = state.getItemSlot(bagId, itemSlot);
-                if (bagItemId == itemId) {
+                if (bagItemId == itemId && balance > 0) {
                     // Found item
                     return (bagSlot, itemSlot, balance);
                 }


chrisfarms (approved):
LGTM

... however there is also some similar find item code that may be suffering from this bug:

* https://github.com/playmint/ds/blob/907c4771e2777029fe2db2dca2482affd115bd18/contracts/src/maps/labyrinth/kinds/buildings/Gates/PrepGate.sol#L51 (think this might be ok due to the way it counts balances)
* https://github.com/playmint/ds/blob/907c4771e2777029fe2db2dca2482affd115bd18/contracts/src/maps/example-gate/Gate/Gate.sol#L52 (exact same code in the example)

but will leave it up to you to decide if there is anything worthy of fixing as part of this PR

5p0rt5BEArD:
> LGTM
> 
> ... however there is also some similar find item code that may be suffering from this bug:
> 
> * https://github.com/playmint/ds/blob/907c4771e2777029fe2db2dca2482affd115bd18/contracts/src/maps/labyrinth/kinds/buildings/Gates/PrepGate.sol#L51
>    (think this might be ok due to the way it counts balances)
> * https://github.com/playmint/ds/blob/907c4771e2777029fe2db2dca2482affd115bd18/contracts/src/maps/example-gate/Gate/Gate.sol#L52
>    (exact same code in the example)
> 
> but will leave it up to you to decide if there is anything worthy of fixing as part of this PR

Nice. I will fix the other places up since it's example code.

hypnoshock:
ah of course *face palm*. It was returning the first slot it found of that type which had a balance of zero! Then afterwards the balance was compared in the hook code and it would revert. So I guess that the code that is used to move items to the wallet was setting the balance to zero but not setting the item type to zero (which is fine but explains what was happening)

hypnoshock (approved):
Nicely fixed!

ldunnplaymint (approved):

5p0rt5BEArD:
Added new issue for doing the same in other example code
#1149

PR #1146, opened 2024-03-11
Duck Burger - Croissant: Moved Display Buildings Down

### LINKED ISSUE #1137: Move dB display buildings to front of area
So you can see scores and time as it fills up

Steedie:
## What
The duck burger display buildings have been moved to the bottom of the arena in the croissant map.
image

## Why
Because the displays on these buildings would previously become hidden when there were Weak Ducks/Burgers placed in front of them.

## How
Edited locations in the croissant map.yaml

Resolves #1137

### DIFF
--- a/contracts/src/maps/croissant/map.yaml
+++ b/contracts/src/maps/croissant/map.yaml
@@ -39,7 +39,7 @@ spec:
 kind: Building
 spec:
   name: Burger Display Building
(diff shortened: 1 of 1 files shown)


chrisfarms (approved):
checked it still counts ducks and counts down ... LGTM

PR #1147, opened 2024-03-11
Item tooltip positioning falling off top of screen

### LINKED ISSUE #1118: item popover pops-over the top of the screen
the item popover (that shows item composition) gets clipped off the top of the screen if item slot is near the top of screen

hypnoshock:
# What

- If an item tooltip is set to be `top` and it falls off the screen, the position will be changed to `bottom`
- Fixed CSS used for 'bottom' position as the pointer and tooltip box were not in the expected place. I think something had bit rotted in the CSS and as we never used bottom positioning we didn't realise it had stopped working

image

# Why

Item tooltips were falling off the top of the screen making them unreadable. This was a problem for the Labyrinth game as item names are important to solve one the puzzle rooms

Fixes #1118

### DIFF
--- a/frontend/src/plugins/tooltip/tooltip.tsx
+++ b/frontend/src/plugins/tooltip/tooltip.tsx
@@ -36,8 +36,6 @@ const Tooltip: React.FC<TooltipProps> = ({ children, delay = 400, direction = 't
         }
 
         const rect = tooltipRef.current.getBoundingClientRect();
-        // console.log('rect:', rect);
-
         if (rect.y < 0 && tooltipDirection === 'top') {
             setTooltipDirection('bottom');
         }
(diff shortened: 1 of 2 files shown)


frontend/src/plugins/tooltip/tooltip.tsx
```diff
@@ -24,11 +26,30 @@ const Tooltip: React.FC<TooltipProps> = ({ children, delay = 400, direction = 't
         setActive(false);
     };
 
+    useEffect(() => {
```
chrisfarms, line 29: this _might_ be one of the rare use cases for `useLayoutEffect` 

https://react.dev/reference/react/useLayoutEffect

... but I certainly can't detect any flicker so it's probably fine

chrisfarms (approved):

ldunnplaymint (approved):

ldunnplaymint (approved):

PR #1150, opened 2024-03-11
Duck Burger: Default frontend time to 3 minutes

### LINKED ISSUE #1138: dB display building default to 3 mins
So it matches the game time
If it's dynamic great. Hardcoded to 3 will do.

Steedie:
## What
Duck Burger Countdown building default display time changed from 1 minute to 3 minutes.
image

## Why
To match it with the actual Duck Burger game time which is set in the solidity code.

## How
By modifying `countdownTotalTime` in `DuckBurgerHQ.js`
```js
const countdownTotalTime = 60000 * 3;
```

## Notes
During kick-off, we thought we may have had to modify all the display building code to take `duration` instead of `endTime` and handle that slightly differently to allow the plugin to set it dynamically. However, it's already able to dynamically set the default time, so no further change is necessary.

Resolves #1138


chrisfarms (approved):
hah I did not see that `else` block that used countdownTotalTime 🤦 

good stuff

PR #1153, opened 2024-03-11
Don't show bags for units that have timed out

### LINKED ISSUE #1144: Inactive (hidden) unit bags are still displayed in the tile panel
image

Though the unit model is hidden due to inactivity, their bag is still displayed on the tile they're on.

hypnoshock:
# What

After a unit has timed out and not on the map, the unit's bag is no longer displayed in the tile info panel

# Why

The unit is dead, we dont' want to see their bags

Fixes #1144


ldunnplaymint (approved):

chrisfarms (approved):
why I actually waited 30mins instead of changing the timeout to be smaller I'll never know ... but confirmed working 😅

PR #1156, opened 2024-03-12
Added 'Facing Direction' to buildings

JackPlaymint:
This PR adds a way to control the rotation of buildings, while also aligning all buildings to a set direction by default.

It adds the concept of 'Facing Direction' which can have a value of either RIGHT or LEFT.
By default, all buildings face to the right of the screen, however this can be overridden in the map manifest on a per-instance basis. For example:
```
kind: Building
spec:
  name: Iron Gate
  location:
    - 12
    - -16
    - 4
  facingDirection: LEFT
``` 
I have added the above example to one of the buildings in the croissant labyrinth game, however I have not converted the gates to door models or applied correct rotations to the other gates.

### DIFF
--- a/cli/src/commands/get.ts
+++ b/cli/src/commands/get.ts
@@ -1,7 +1,7 @@
 import { z } from 'zod';
 import { WorldStateFragment, getCoords } from '@downstream/core';
 import { pipe, take, toPromise } from 'wonka';
-import { BiomeTypes, Manifest, Slot } from '../utils/manifest';
+import { BiomeTypes, FacingDirectionTypes, Manifest, Slot } from '../utils/manifest';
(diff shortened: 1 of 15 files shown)


chrisfarms (commented):
code looks good ... spun it up and it does what it says 👍 

couple of bits needed:

* It looks like you need to add `facingDirection: 'LEFT'` to the bit in the tile-fabricator that builds manifests since the manifest definition changed.
* It looks like you will need to update all uses of `DEV_SPAWN_BUILDING` in `LabyrinthCore.sol`

hypnoshock (approved):
I tested the tile fabricator and can see `facingDirection` come through as either `LEFT` or `RIGHT` so it appears to like the TypeScript syntax I threw at it

PR #1157, opened 2024-03-12
Labyrinth reset tower

hypnoshock:
# What

Added a reset tower which will call the `reset` function on the `LabyrinthCore` contract. 

image

- This building is both present in the standalone and croissant maps
- A password must be supplied to reset
- The reset code has now been made a public function on the core building which can only be called by either the core building itself or the reset tower

# Why

Adds the ability to reset the Labyrinth without having to walk to the core building and supply 4 hearts

# Notes

If we want to change the password run a password (all lower case) through a keccak256 hash (can be done here: https://emn178.github.io/online-tools/keccak_256.html) and then update the `getPasswordHash` function in the `ResetTower` contract

```
    function getPasswordHash() internal pure returns (bytes32) {
        return 0x1f1fc8e191099524ede873ed4b43087e02baaa5f39243fdd808f60582dc7702e;
    }
```

### DIFF
--- a/contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/LabyrinthCore.sol
+++ b/contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/LabyrinthCore.sol
@@ -11,6 +11,8 @@ import {ILabyrinthCore} from "./ILabyrinthCore.sol";
 using Schema for State;
 
 contract LabyrinthCore is ILabyrinthCore, BuildingKind {
+    bytes24 constant _RESET_TOWER = 0xbe92755c0000000000000000f1e6d10a0000000000000004;
+
     // for respawn
     bytes24 constant _LARGE_ROCKS = 0xbe92755c0000000000000000d4c1c6880000000000000001;
     bytes24 constant _CRUSHER = 0xbe92755c0000000000000000e92c4edd0000000000000001;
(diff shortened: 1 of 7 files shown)


contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/LabyrinthCore.sol
```diff
@@ -20,10 +21,21 @@ contract LabyrinthCore is BuildingKind {
     function use(Game ds, bytes24 buildingInstance, bytes24, /*actor*/ bytes memory /*payload*/ ) public override {
         // Crafting the Playtest Pass
         ds.getDispatcher().dispatch(abi.encodeCall(Actions.CRAFT, (buildingInstance)));
+        reset(ds, buildingInstance, buildingInstance);
```
chrisfarms, line 24: non-blocking: probably doesn't need to be doing a full CALL to itself, could be calling an internal func shared by both this and the public reset

contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/ResetTower.sol
```diff
@@ -0,0 +1,53 @@
...
+import {Schema, Kind, Node} from "@ds/schema/Schema.sol";
+import {Actions} from "@ds/actions/Actions.sol";
+import {BuildingKind} from "@ds/ext/BuildingKind.sol";
+import {ILabyrinthCore} from "./ILabyrinthCore.sol";
+
+using Schema for State;
+
+contract ResetTower is BuildingKind {
+    function getPasswordHash() internal pure returns (bytes32) {
+        return 0x1f1fc8e191099524ede873ed4b43087e02baaa5f39243fdd808f60582dc7702e;
```
chrisfarms, line 16: 🤫

contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/ResetTower.js
```diff
@@ -0,0 +1,163 @@
...
+    );
+}
+
+function logState(state) {
+    console.log("State sent to pluging:", state);
+}
+
+const friendlyPlayerAddresses = [
+    // 0x402462EefC217bf2cf4E6814395E1b61EA4c43F7
+];
```
chrisfarms, line 125: **nit:** lots of these funcs seem unused, should we remove them?

contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/ResetTower.yaml
```diff
@@ -0,0 +1,25 @@
...
+  inputs:
+    - name: Red Goo
+      quantity: 20
+    - name: Green Goo
+      quantity: 10
+    - name: Blue Goo
+      quantity: 2
+  outputs:
+    - name: Bronze Key
+      quantity: 1
```
chrisfarms, line 25: **nit:** these can be removed since it's `custom` right?

chrisfarms (approved):
set the password, and made it do the thing ... I'm probably not best placed to know if it did the _right_ thing, so maybe @5p0rt5BEArD can take a look ... but LGTM

couple of comments, but mostly just some tidy up of dead code

hypnoshock:
Agreed, I'll fix these things before merge

contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/LabyrinthCore.sol
hypnoshock: Fixed. Now calls an internal `doReset` function. The public reset does the check to see if it's the reset building and if that passes calls the same internal `doReset`

contracts/src/maps/labyrinth/kinds/buildings/LabyrinthCore/ResetTower.yaml
hypnoshock: Good point, all gone!

Issue #1091, opened 2024-02-28
cannot drag item from a public slot directly to wallet-inventory


chrisfarms:
try moving an item from an extractor/craft slot straight into your wallet inventory .... it will fail, it shouldn't

Issue #1066, opened 2024-02-20
decide: how or if wallet-inventory to unit-inventory item transfers are restricted


chrisfarms:
directions:

* could be unrestricted
* could be only possible at a location (like center of map)
* could be dictated by the item contract if the item is allowed to be transfered (builder controlled)

PR #1176, opened 2024-03-18
allow xfer item from building-bag to wallet-bag

chrisfarms:
lets you export items directly from building-attached bags like input/output slots to player wallet inventory.

we currently allow xfer from wallet inventory TO building-attached bags, so this makes it consistent.

HOWEVER; both of these things are probably not quite what we want as they do allow for teleporting items from long distance bags in some scenerios. We have an issue in the backlog about deciding if/how this kind of transfer should work. So at this time this is more about making it not appear broken than about deciding on how to enforce location-aware item transfers

fixes: #1091

### DIFF
--- a/contracts/src/rules/InventoryRule.sol
+++ b/contracts/src/rules/InventoryRule.sol
@@ -59,8 +59,10 @@ contract InventoryRule is Rule {
         address toAddress,
         uint64 qty
     ) private {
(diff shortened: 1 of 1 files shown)


5p0rt5BEArD (approved):
Looks good and you've explained we have a future issue to set the rules properly (issue #1066)

Issue #1163, opened 2024-03-18
Tile Fabricator to read current state of map


ldunnplaymint:
I want to be able to open the tile tabricator and with the press of a button, import the current map of the associated hexwood/URL

PR #1177, opened 2024-03-18
load the current tiles/kinds into tile-fab on load

chrisfarms:
### what

* When the tile-fabricator first loads it will now populate the grid with the Tiles, Buildings, and BuildingKinds from the game state.
* Reset will reload the page to reset back to the game state.
* Clear will wipe all the tiles, giving a blank canvas.
* Removed the diameter slider, it just gets in the way of loaded maps
* updated `ds apply -R` to find .yml files too (since tilefab exports as .yml in chrome) 

### why

Showing the current map is more intuitive, and makes it easier to deploy a BuildingKind with building-fab, then jump to tile-fab to place it.

resolves: #1163 

### notes

* state is only ever loaded on first load, it will not update based on live changes ... this avoids the situation where updates destroy or corrupt your creation while creating it.
* it's not really clear what should happen if you import on top of the state-loaded map ... currently it is addative, so you would need to press clear then import to get a clean import... this is so that you can still load additional things

### todo

* [x] fixup the nasty `any` types
* [x] fixup the broken preview tile not displaying
* [x] fixup the broken preview buildings not displaying
* [x] fixup the .yml extensions (or fix cli to accept them)

### DIFF
--- a/cli/src/commands/apply.ts
+++ b/cli/src/commands/apply.ts
@@ -23,7 +23,7 @@ export const getManifestFilenames = (filename: string, isRecursive: boolean): st
             throw new Error(`${filename} is a directory. use --recursive to apply all manifests in a directory`);
         }
         // must be posix path, even on windows
-        const globPath = path.join(filename, '**/*.yaml').replace(/\\/g, '/');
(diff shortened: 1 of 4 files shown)


5p0rt5BEArD (approved):
Looks good and works so far

Issue #1077, opened 2024-02-23
As an aa-hacker I want an intuitive way to browse and deploy example plugins and maps


5p0rt5BEArD:
We currently have example-plugins/ with some building kinds demonstrating specific features and a maps folder with some complete experiences as well some specific feature demos.

There are some symlinks between the two to reduce repeating the definition of common building kinds.

Let's refactor the current structure to be clear as possible. Goals are
- Once you've found one you've found them all
- Easy to browse to some functionality you'd like to fork
- Easy to deploy "experience" maps
- Easy to deploy "feature example" maps
- Reduced duplication of kinds when used on multiple maps
- Documented (cheatsheet/how-to-docs/inline)
- Consistent naming

A possible example arrangement is

/maps/
-/basicfactory/
--/basicfactory-readme.md
--/basicfactory-map.yaml
--/basicfactory-kinds/
---/basicfactory.js
---/basicfactory.sol
---/basicfactory.yaml
-/blockers/
--/blockers-readme.md  (explain blocker category)
--/blockers-map.yaml
--/blockers-kinds.yaml (no sub dir because there's no code?)
-/cocktailhut/
--/cocktailhut-readme.md  (explain item kinds, fetch and always actiive)
--/cocktailhut-map.yaml
--/cocktailhut-kinds/
---/cocktail.js
---/cocktailhut.js
---/cocktailhut.sol
---/cocktailhut.yaml
-/croissant/
--/croissant-readme.md
--/croissant-map.yaml/
--/croissant-kinds/
---/ (symlinks to other kinds folders)
-/duckburger/
--/duckburger-readme.md
--/duckburger-map.yaml
--/duckburger-kinds/
---/ (current contents of DB with symlinks to basicfactory)
-/gate/
--/gate-readme.md  (explain gates)
--/gate-map.yaml
--/gate-kinds/
---/gate.js
---/gate.sol
---/gate-items.yaml
---/gate.yaml

chrisfarms:
some more notes from chatting out of band:

* symlinks are too annoying for windows users ... let's try to avoid
* duplication is preferred over symlinking to keep each map folder isolated (we will accept the debt of harder maintainence of the complex maps in exchange for less surprizing behaviour and simplier to grok self-contained map folders)
* we think avoiding nesting folders within maps will make it easier to use with the tile fabricator (which is unable to load a recursive folder structure)
* we should use more meaningful names for yaml manifests so that you can easily find things:
[...]

PR #1180, opened 2024-03-19
Tutorial room 3

JackPlaymint:
This PR adds a third room to the tutorial.
It demonstrates moving your unit, constructing a building, crafting an item, using quests and potentially adding an item to your player wallet.

The quests are set up to guide the player through the moving and construction tasks and demonstrate a range of quest types.


ldunnplaymint:
Screenshot 2024-03-19 at 16 07 20

Whenever I run this, I just get an empty map: Also the tiles don't extend far enough so I can spawn and walk to the island.

JackPlaymint:
> Screenshot 2024-03-19 at 16 07 20
> Whenever I run this, I just get an empty map: Also the tiles don't extend far enough so I can spawn and walk to the island.

Do you get a quest? You should just have an empty map with a quest.

I didn't include the centre island as I thought that would be there by default and these spokes would be added after that initial centre island was made.

ldunnplaymint:
Screenshot 2024-03-19 at 16 13 59

My bad, I do get the quest but I am unable to move the Unit without those two extra tiles. Likely worth chucking them in, I think if the HUB yaml has them too, one set just gets dropped anyway.

JackPlaymint:
> I added the tiles locally and I am happy with the room. Once they go in, will tick and merge.

Awesome, just pushed those up now :)

contracts/src/maps/tutorial-room-3/building-kinds/CastleOfTreasure.js
```diff
@@ -0,0 +1,141 @@
...
+
+function unitIsBuildingOwner(mobileUnit, selectedBuilding) {
+    //console.log('unit owner id:',  mobileUnit?.owner?.id, 'building owner id:', selectedBuilding?.owner?.id);
+    return mobileUnit?.owner?.id && mobileUnit?.owner?.id === selectedBuilding?.owner?.id;
+}
+
+function unitIsBuildingAuthor(mobileUnit, selectedBuilding) {
+    //console.log('unit owner id:',  mobileUnit?.owner?.id, 'building author id:', selectedBuilding?.kind?.owner?.id);
+    return mobileUnit?.owner?.id && mobileUnit?.owner?.id === selectedBuilding?.kind?.owner?.id;
+}
```
chrisfarms, line 133: think we should remove all the unused code in these plugins to keep the examples concise

chrisfarms (commented):
in @Steedie's PR (and related to #1077 ) .. we have said we are going to avoid nesting folding within each map from now on, as it makes working with tile-fab harder and easier to see what is in each example

So could we ensure there are no folders inside tutorial-room-3 (sorry for the faff!)

JackPlaymint:
> in @Steedie's PR (and related to #1077 ) .. we have said we are going to avoid nesting folding within each map from now on, as it makes working with tile-fab harder and easier to see what is in each example
> 
> So could we ensure there are no folders inside tutorial-room-3 (sorry for the faff!)

Oh yeah! No worries

Issue #1167, opened 2024-03-18
Remove the docs webpage


ldunnplaymint:
Downstream-URL/docs needs to go, just remove the deployment

PR #1181, opened 2024-03-19
remove /docs page

chrisfarms:
we are doing some docs/guides in github and the older istanbul docs are confusing to have around at the moment

leaves the markdown bits intact for now as we may be porting some bits from them

resolves: #1167


Steedie (approved):
👍

PR #1182, opened 2024-03-19
option to deploy to redstone

chrisfarms:
### what

adds option to the Github Action deployment workflow to deploy to Redstone Holesky

Screenshot 2024-03-19 at 17 56 26

### why

we think it's a good location for some hackathon antics

resolves: #1166 

### notes

the "reset" action will not work for redstone deployments, since the reset action works by destroying the chain, but that is not an option here

### todo

* [x] check the old anvil deployment still works
* [x] it's not currently doing the ds apply for map

### DIFF
--- a/.github/workflows/build-hexwood.yml
+++ b/.github/workflows/build-hexwood.yml
@@ -38,6 +38,13 @@ on:
           - croissant
           - example-gate
           - labyrinth
+      SELECTED_CHAIN:
+        type: choice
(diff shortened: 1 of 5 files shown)


ldunnplaymint (approved):
I can see the redstone deploy, I can confirm it is on redstone and I can infer the deploy process worked because it is there. Probably want a code review as well?

PR #1183, opened 2024-03-20
Tutorial Room 1

JackPlaymint:
This PR adds tutorial room one. It contains some blockers and a basic factory made using the Tile Fabricator.
Most of this island will need to be shown off in the video tutorial as it mostly focuses on the tile and building fabricator, so the island itself is pretty basic.


contracts/src/maps/tutorial-room-1/room-1-map.yaml
chrisfarms, line 1: think we should call this file "tiles.yaml" (in an attempt to disambiguate between us calling the entire folder a "map" and calling tiles a "map")

ldunnplaymint (approved):

contracts/src/maps/tutorial-room-1/room-1-map.yaml
JackPlaymint: Sure thing, I changed it for room 1 and  room 3 as well.

PR #1194, opened 2024-03-22
Structure Refactoring

### LINKED ISSUE #1077: As an aa-hacker I want an intuitive way to browse and deploy example plugins and maps

Steedie:
## What
During a kick-off with @chrisfarms @hypnoshock @5p0rt5BEArD & @ldunnplaymint, we decided on a format for the refactoring task: #1077

We decided on these rules:
- We want to "flatten" our map folder structure
- Use PascalCase
- We want a consistent naming structure
  - Anything with a location in `Locations.yaml` (Or `Buildings.yaml`, `Bags.yaml`, `Tiles.yaml` where possible)
  - Any `Kinds` in their own `.yaml`
- No more "Common" folder
  - Make a new map called `blocker-land` & `extractor-land` so there's a clear example of them being used
- No more example-plugins folder
  - These should be made into their own maps in the `maps` folder
- No use of symlinks

We also decided that we would only apply these changes to our own maps, so I haven't touched Tonk Attack, Hexcraft, or Labyrinth for now.

### Summary
- [x] Flat folder structure in our maps
- [x] Update .yaml naming conventions
- [x] Extract /common folder
- [x] Extract /example-plugins
- [x] Remove symlinks
- [x] Remove /common folder
- [x] Remove /example-plugins
  - (DuckBurger has been left in because the performance-test currently relies on it)

### common -> maps
- [x] `common/Decorations` -> `maps/example-blockers`
- [x] `common/extractors` -> `maps/example-extractors`
- ~~_`/common` still exists because symlinks (which will be getting removed) are still using them_~~

### example-plugins -> maps
- [x] `example-plugins/BasicFactory` -> `maps/basic-factory`
  - ~~_`BasicFactory` is being used by some symlinks (to be removed)_~~
- [x] `example-plugins/Cocktails` -> `maps/cocktail-hut`
- [x] `example-plugins/DuckBurger` -> `maps/duck-burger`
  - _`DuckBurger` is being used by the performance-test_
- [x] `example-plugins/StateStorage` -> `state-storage`

Resolves #1077


contracts/src/maps/example-blockers/Locations.yaml
```diff
@@ -0,0 +1,162 @@
+
```
chrisfarms, line 1: think you are missing a 0,0,0 Tile in this list

Screenshot 2024-03-25 at 14 29 35

chrisfarms (commented):
Good stuff, I'm sure that was a tedious task

some thoughts, but nothing blocking:

* 🤔 "quest-map" is the only map with "map" in the name, and it seems redundent. Maybe this should be called "istabul" (since thats what we call it) ... should get @5p0rt5BEArD feedback their though
* 🤔 some maps are prefixed with "example-" and others (that are arguably examples) like state-storage are not ... should we be more consistent here (ie never prefix xor always prefix)
* 🤔 There's a mix of "-" (hyphen) vs "_" (underscore) used in folder names ... this might have been there previously but would be nice to standardize on the same thing everywhere (hyphen probably since they are winning)
* 🤔 There's still some `map.yaml` files laying around ... I think because we said leave the "games" alone, but maybe worth renaming these to Location.yaml or Tiles.yaml to avoid the "map is the tiles" vs "map is the whole map with stuff" confusion.

chrisfarms (commented):
Looks like there are still files in [contracts/src/example-plugins](https://github.com/playmint/ds/tree/yaml-refactor/contracts/src/example-plugins/) (a copy of duck burger). I think we wanted to remove this dir.

...which will also mean removing the `ds apply` line in .devstartup.js that is currently applying the example-plugins during `make dev`

Steedie:
> Looks like there are still files in [contracts/src/example-plugins](https://github.com/playmint/ds/tree/yaml-refactor/contracts/src/example-plugins/) (a copy of duck burger). I think we wanted to remove this dir.
> 
> ...which will also mean removing the `ds apply` line in .devstartup.js that is currently applying the example-plugins during `make dev`

It's also being used by [scripts/generate-perf-test-map.js](https://github.com/playmint/ds/blob/main/scripts/generate-perf-test-map.js), not completely sure if changing the dir it's using their will break the performance-test too, I'll give that a go now.

contracts/src/maps/example-blockers/Locations.yaml
Steedie: Solved in latest commit 🙏

Steedie:
> Looks like there are still files in [contracts/src/example-plugins](https://github.com/playmint/ds/tree/yaml-refactor/contracts/src/example-plugins/) (a copy of duck burger). I think we wanted to remove this dir.
> 
> ...which will also mean removing the `ds apply` line in .devstartup.js that is currently applying the example-plugins during `make dev`

These changes are now implemented in the latest commit. Also, fixed the performance-test related stuff since it previously relied on example-plugins/DuckBurger

chrisfarms (commented):
I think we said we would add a `README.md` to every map folder ... nothing big, just a one liner description is enough for now .... ie:

`./contracts/src/maps/example-blockers/README.md` might be:

```
# Blockers

A small map showing the available building models for blocker category of buildings
```

Steedie:
> Good stuff, I'm sure that was a tedious task
> 
> some thoughts, but nothing blocking:
> .......

1. Waiting for feedback from @5p0rt5BEArD
2. Solved in [commit](https://github.com/playmint/ds/pull/1194/commits/537e1641a017bb697e16db11b399e942bc67ee37)
3. Solved in [commit](https://github.com/playmint/ds/pull/1194/commits/d5053aaeaa2370442583e173fcf07c2231a0e24a)
4. Solved in [commit](https://github.com/playmint/ds/pull/1194/commits/9885ccda8387132e534e8fec0a8d1e529f2395c8)

contracts/src/maps/state-storage/StateStorage.js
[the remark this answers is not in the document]
Steedie: Fixed plugin error: https://github.com/playmint/ds/pull/1194/commits/eab1824e8fcccaf98ce4ef76a1b2a174fe61d039#diff-115acbc2c5bccb5e889b1e539b432a9cb328e058abec1548d37c82f3deaaea7cR17-R18

Steedie:
> > I think we said we would add a `README.md` to every map folder ... nothing big, just a one liner description is enough for now .... ie:
> > `./contracts/src/maps/example-blockers/README.md` might be:
> > ```
> > # Blockers
> > 
> > A small map showing the available building models for blocker category of buildings
> > ```
> 
> Ah yep I missed this. I'll work on these next.

SImple READMEs for each map have been added now ([commit](https://github.com/playmint/ds/pull/1194/commits/9c544fe37677358d30d74b5d3a3820690be6849e))

chrisfarms (approved):

PR #1209, opened 2024-03-25
Tutorial room 3 readme

JackPlaymint:
Here's a readme for the third tutorial room.

### DIFF
--- a/contracts/src/maps/tutorial-room-3/Readme.md
+++ b/contracts/src/maps/tutorial-room-3/Readme.md
@@ -4,7 +4,7 @@
 We will follow the steps below to create a series of quests which you will use to guide the player to complete various tasks.
 
 ## 1. Preparation
-- Before starting, ensure that you have deployed an instance of Downstream locally using Docker. You can find instructions on how to do that in [Tutorial 1 here](../tutorial-room-1/Readme.md).
+- Before starting, ensure that you have deployed an instance of Downstream locally using Docker. You can find instructions on how to do that in [Tutorial 1 here](../tutorial-room-1/README.md).
 - You should also copy the CastleOfTreasure .yaml, .sol and .js files out of the tutorial room 3 folder and paste them into your working folder.
 
 
--- a/contracts/src/maps/tutorial-room-3/Readme.md
+++ b/contracts/src/maps/tutorial-room-3/Readme.md
@@ -12,28 +12,37 @@ We will follow the steps below to create a series of quests which you will use t
 
 We will begin by creating our first quest, which will ask the player to move to a certain location. Create a new file called “00_MoveToLocation.yaml” and open it in your IDE.
 Begin by adding the following lines to define that the code in this file will be representing a Quest:
+
 ```yaml
 ---
 kind: Quest
 ```
+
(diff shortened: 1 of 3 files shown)


contracts/src/maps/tutorial-room-3/Readme.md
```diff
@@ -0,0 +1,132 @@
...
+```
+Next you should add information about the Quest itself:
+```yaml
+spec:
+    name: Move to the South-East Island
+    description: "Click 'Move', then click on a tile. Then click 'Confirm' to move your Unit to that location."
+```
+You can then define what tasks the player must do to complete this quest. For this first quest, we want the player to move to a location, so we’ll add a “coord” type task:
+```yaml
+tasks:
```
chrisfarms, line 31: think this needs indenting

```suggestion
    tasks:
```

contracts/src/maps/tutorial-room-3/Readme.md
```diff
@@ -0,0 +1,132 @@
...
+ name: Construct a Castle of Treasure
+ description: "Click 'Build', then click on a tile. Then select 'Castle of Treasure' from the list of buildings and click 'Confirm'"
+ tasks:
+   - kind: construct
+     name: Construct a Castle of Treasure.
+     buildingKind: Castle of Treasure
+ next: ["Create Some Treasure"]
+```
+Note that in this quest we are using a “construct” task type instead of a “coord” task type as we did in the previous example, and we reference what building we want the player to build by name.
+It’s also important that the name of this quest matches the name we provided to the “next” property in the previous quest’s code.
```
chrisfarms, line 72:
```suggestion
It’s important that the name of this quest matches the name we provided to the “next” property in the previous quest’s code.
```

contracts/src/maps/tutorial-room-3/Readme.md
```diff
@@ -0,0 +1,132 @@
...
+ index: 0
+```
+
+# 4. Deploy your new Quests
+
+We will use the Downstream CLI to deploy our newly created building to our local Downstream instance.
+
+Run the following in a terminal in the folder with your exported building source (refer to tutorial 1 for more information).
+```
+ds apply -n local -k <private-key>> -R -f ./
```
chrisfarms, line 110:
```suggestion
ds apply -n local -k <private-key> -R -f ./
```

contracts/src/maps/tutorial-room-3/Readme.md
```diff
@@ -0,0 +1,132 @@
...
+```
+You should see an output like this:
+
+<img src="./readme-images/deployOutput.png" width=300>
+
+After that, you should see your first quest in the UI in your local instance of downstream. After completing each quest you should then see the next until you have completed them all.
```
chrisfarms, line 116: Might be nice to have a screenshot of the quest UI here (or somewhere) to tie together some of the text fields with what you expect to see on the screen

chrisfarms (approved):
some minor optional comments but think it's looking good.

There's probably some opportunity to cut back on the more conversational-flow bits, if we needed to, but sounds good as it is.

chrisfarms (approved):

PR #1215, opened 2024-03-27
tutorial room 5 readme

hypnoshock:
# What

The readme for 5 turned into a full tutorial. The first 4 steps are the same as tutorial one's so maybe I could summarise steps 1 to 4 as one step to make it shorter


contracts/src/maps/tutorial-room-5/README.md
```diff
@@ -1,76 +1,667 @@
...
-    -   The `startBlock` and `endBlock` properties which are used to drive the `Coundown` buildings which are set when the countdown is started
+We will follow the steps below to create a downstream map with your own tile layout, a building that can incrememnt an on-chain counter, two buildings that display the counter value, a building that can start a countdown timer and two buildings which will display the countdown.
 
-## Displaying a number on a Counter building
+Once complete, you will have used the core Downstream tools to create a map that looks like this when played:
 
-The `CounterHQ` building drives the display of the `Counter` buildings by finding all Counter buildings within a 2 tile radius of the HQ and setting the number to be displayed by setting the `map` property on the building's output object
+<img src="./readme-images/screenshot.png" width=300>
 
+# Prerequisites
```
chrisfarms, line 11: this title and all subsequent at this level should be H2... + bump any H2 -> H3 where sub-sub have been used

```suggestion
## Prerequisites
```

contracts/src/maps/tutorial-room-5/README.md
chrisfarms, line 1: I think we could probably condense steps 1 - 4 into a single step that assumes prior knowledge from tutorial-1, since this is what the room-3 does.

**but** it reads _really_ well as-is so let's not make work for ourselves here today, and maybe we can come back another day to do a consistency pass over all the tutorials.

contracts/src/maps/tutorial-room-5/README.md
```diff
@@ -1,76 +1,667 @@
...
-        value: `${count % 100}`,
-    })),
+
+After some time (this could take up to 5 minutes), you should see "ready" in the terminal output:
+
+    contracts-1  | +-------+
+    contracts-1  | | ready |
+    contracts-1  | +-------+
+
+You can now open a web browser and navigate to [http://localhost:3000] to load the game.
```
chrisfarms, line 33: should this be a link...

```suggestion
You can now open a web browser and navigate to [http://localhost:3000](http://localhost:3000) to load the game.
```

contracts/src/maps/tutorial-room-5/README.md
```diff
@@ -1,76 +1,667 @@
...
+          quantity: 25
+        - name: Blue Goo
+          quantity: 25
+        - name: Red Goo
+          quantity: 25
+```
+
+## Finding all counter buildings within a 2 tile radius
+
+To make this easier we can use the following helper functions which can be pasted at the end of `CounterHQ.js`
```
chrisfarms, line 346: this might need a _little_ more info on what they are doing as they are quite weird funcs

```suggestion
The raw tile coordinates need decoding a bit to calculate distances between them. To make this easier we can use the following helper functions, which can be pasted at the end of `CounterHQ.js`
```

... or something

contracts/src/maps/tutorial-room-5/README.md
```diff
@@ -1,76 +1,667 @@
...
+    const payload = ds.encodeCall("function startTimer(uint256)", [endBlock]);
+
+    ds.dispatch({
+        name: "BUILDING_USE",
+        args: [countdownHQ.id, mobileUnit.id, payload],
+    });
+};
+```
+
+We are calling this function from the button actions which were already defined in the skeleton code that we pasted earlier.
```
chrisfarms, line 643: maybe needs making clear that this is like an event listener

```suggestion
This function is called from our button's `action` handler, which we already defined in the skeleton code earlier.
```

chrisfarms (commented):
I think it reads nicely.

There's maybe a few opportunities totranslate bits into "You ..." rather than "We ..." to highlight more that the reader is doing most of the work ... but I don't think it would add enough value to warrant rewording.

few comments inline

contracts/src/maps/tutorial-room-5/README.md
```diff
@@ -1,76 +1,667 @@
...
+            text: "Start 30 second timer",
+            type: "action",
+            action: () => startTimer(30),
+        },
+        {
+            text: "Start 1 minute timer",
+            type: "action",
+            action: () => startTimer(60),
+        },
+    ],
```
chrisfarms, line 657: it's not immediately obviously where this bit is, _might_ benefit from some snipped context like, but maybe that's just me:

````suggestion
```js
return {
   ...
    buttons: [
        {
            text: "Start 30 second timer",
            type: "action",
            action: () => startTimer(30),
        },
        {
            text: "Start 1 minute timer",
            type: "action",
            action: () => startTimer(60),
        },
    ],
    ....
};
````

ldunnplaymint (approved):

PR #1238, opened 2024-04-02
q,r,s action bounds

### LINKED ISSUE #1200: cap spawning Tiles/Buildings/Bags within a location limit
* we want to loosely cap max amount of stuff by capping where it can spawn
* the tile-fab is already capped ... probably use the same capping size but enforced
* move actions should not allow moving outside the cap
* construct/dev_spawn actions should not allow spawning outside the cap

Steedie:
## What
This PR stops people from doing actions outside the bounds of the map, which we've defined as a radius of 40.
This is happening on both the contract side and the ds-cli applier.

## Why
Before this change, someone could infinitely apply bags, tiles, and buildings, and could technically move a unit to any position.

## How
New utils files containing `MAX_RANGE` (radius) - to determine whether the q,r,s coordinates are within the defined bounds:
- `ds\contracts\src\utils\Bounds.sol`
- `ds\cli\src\utils\bounds.ts`
used by below as needed:

### Contract-side:
- [x] DEV_SPAWN_TILE
- [x] DEV_SPAWN_BUILDING
- [x] DEV_SPAWN_BAG
- [x] MOVE_MOBILE_UNIT
- [x] SCOUT_MOBILE_UNIT

### ds-cli apply
- [x] Tiles
- [x] Buildings
- [x] Bags

Resolves #1200


ldunnplaymint (approved):

hypnoshock (approved):
Looks good to me, I don't have any comments!

PR #1242, opened 2024-04-03
Dev Destroy Actions (WIP)

### LINKED ISSUE #1214: Allow applying "biome" in DEV_SPAWN_TILE
A new ```ds delete``` CLI command that takes the same params as apply but instead of applying the manifest, it attempts to delete the entities that are passed via the manifests. This task is just about tiles, building instances and bags.

If a tile is deleted and a unit is sitting on it, move it to 0,0,0.

There will be DEV_DESTROY_[ENTITY] actions that will match the current DEV_SPAWN actions.

Destroying a tile also destroys the buildings and bags on that tile.

Steedie:
## What
implements `ds destroy`. Behaves similar to ds apply, but instead, it results in removing the node from state.
- DEV_DESTROY_TILE
- DEV_DESTROY_BUILDING
- DEV_DESTROY_BAG

## Notes
- Changes have also been made to `ds\core\src\abi\Actions.ts`, which haven't been pushed to the PR

### Current problem
_(Currently testing destroying tiles)_
When I run:
```solidity
state.remove(Rel.Biome.selector, 0x0, node);
```
It seems to change the tile to `UNDISCOVERED`, but it's still rendered and doesn't remove it from the world. (I get this error when I try to move to the removed tile: "dispatch fail: failed to call: execution reverted: NoMoveToUndiscovered")

Still looking into this ^ WIP...

Resolves #1214

### DIFF
--- a/cli/src/utils/destroyer.ts
+++ b/cli/src/utils/destroyer.ts
@@ -1,0 +1,166 @@
+import { CogAction, NodeSelectors } from '@downstream/core';
+import { BuildingKindFragment } from '@downstream/core/src/gql/graphql';
+import { id as keccak256UTF8, solidityPacked } from 'ethers';
+import { z } from 'zod';
+import {
+    BuildingCategoryEnum,
+    BuildingCategoryEnumVals,
+    ManifestDocument,
+} from '../utils/manifest';
+import { isInBounds } from '../utils/bounds';
+
+export const encodeTileID = ({ q, r, s }: { q: number; r: number; s: number }) => {
+    return solidityPacked(
+        ['bytes4', 'uint96', 'int16', 'int16', 'int16', 'int16'],
+        [NodeSelectors.Tile, 0, 0, q, r, s]
+    );
+};
+
+export const encodeBagID = ({ q, r, s }: { q: number; r: number; s: number }) => {
+    return solidityPacked(['bytes4', 'uint96', 'int16', 'int16', 'int16', 'int16'], [NodeSelectors.Bag, 0, 0, q, r, s]);
+};
+
+const encodeBuildingKindID = ({ name, category }) => {
+    const id = Number(BigInt.asUintN(32, BigInt(keccak256UTF8(`building/${name}`))));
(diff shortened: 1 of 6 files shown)


hypnoshock:
> It seems to change the tile to UNDISCOVERED, but it's still rendered 

My guess is that even though it's `UNDISCOVERED` the tile still has an edge to a location node. On the frontend I'm doubt we are checking bioms at all, we probably just render all tiles that are returned by the query and that probably includes all with a location.

cli/src/utils/destroyer.ts
```diff
@@ -0,0 +1,166 @@
...
+import { id as keccak256UTF8, solidityPacked } from 'ethers';
+import { z } from 'zod';
+import {
+    BuildingCategoryEnum,
+    BuildingCategoryEnumVals,
+    ManifestDocument,
+} from '../utils/manifest';
+import { isInBounds } from '../utils/bounds';
+
+export const encodeTileID = ({ q, r, s }: { q: number; r: number; s: number }) => {
```
hypnoshock, line 12: I suppose all these helper functions here are exactly the same as the ones in the applier. Probably a good idea to move them to a helper file and import them in where they are used as it's a lot to repeat. Not a huge deal though

contracts/src/schema/Schema.sol
```diff
@@ -288,6 +288,11 @@ library Schema {
         return state.set(Rel.Biome.selector, 0x0, node, 0x0, uint64(biome));
     }
 
+    function destroyTile(State state, bytes24 node) internal {
+        // return state.set(Rel.Biome.selector, 0x0, node, 0x0, uint64(biome));
+        return state.remove(Rel.Biome.selector, 0x0, node);
```
hypnoshock, line 293: The fact you're getting that revert message when trying to move a Unit to a deleted tile proves this is working 👍

ldunnplaymint (approved):

hypnoshock (approved):
Looks good to me. Comments I had in during the draft PR have all been addressed, nice one!

PR #1244, opened 2024-04-04
Fourth (zone) Dimension

JackPlaymint:
This PR Adds a fourth dimension to just about everywhere that position coordinates are used. The extra value is the ID for the zone that the coordinates are in.
So coords are now generally: z (zone) ,q,r,s

This is a big ol' update that touches a ton of stuff. Everything seems to be working still, but I'm unsure:
a) if I've broken something somewhere and not realised it
b) if the zone ID will actually work when we start using the zone ID.

For now the zone ID value is hard coded as 0 in applier.ts (temporaryZoneConstant).

### DIFF
--- a/frontend/src/plugins/action-context-panel/index.tsx
+++ b/frontend/src/plugins/action-context-panel/index.tsx
@@ -166,7 +166,7 @@ const Construct: FunctionComponent<ConstructProps> = ({
         }, new Map() as SlotMap);
 
     const targetBuildingId = constructionCoords
-        ? getBuildingId(constructionCoords.q, constructionCoords.r, constructionCoords.s)
+        ? getBuildingId(constructionCoords.z, constructionCoords.q, constructionCoords.r, constructionCoords.s)
         : undefined;
     const targetBagId = targetBuildingId ? getBagId(targetBuildingId) : undefined;
     const targetEquipKey = 0;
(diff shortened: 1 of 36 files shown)


frontend/src/plugins/action-context-panel/index.tsx
```diff
@@ -291,7 +291,7 @@ const Construct: FunctionComponent<ConstructProps> = ({
                     return [
                         {
                             name: 'MOVE_MOBILE_UNIT',
-                            args: [mobileUnitKey, q, r, s],
+                            args: [mobileUnitKey, _zone, q, r, s],
```
chrisfarms, line 294: the underscore in `_zone` was to hint to the linter that we know it's an unused var ... since it's used now, we can remove `_`  

```suggestion
                            args: [mobileUnitKey, zone, q, r, s],
```

chrisfarms (commented):
👏  only had a quick skim will look again tomorrow, will probably need us all to have a review pass since it's a bigun!

I think my main comment is that I believe there are now many cases where rules are checking things like distance between two tiles ... or checking if a unit/bag/building is adjacent to something ... and these things are now not taking into account the fact that the unit may be on another plane of existance 👻  .... so we should try and look out for these cases

chrisfarms:
> I think my main comment is that I believe there are now many cases where rules are checking things like distance between two tiles ...

I have spun out a new issue for fixing up the Rules/Actions so we can keep this PR just about the coords stuff https://github.com/playmint/ds/issues/1245

ldunnplaymint (commented):
I have tested the quest map and done everything the game has to offer (In-game and CLI stuff)

ldunnplaymint (approved):

hypnoshock:
I've done a speedrun of the labyrinth and can confirm it reset as expected with buildings etc spawned back in the correct places

hypnoshock (approved):
Thumbs up from me. What a mission Jack, nice work!

PR #1246, opened 2024-04-05
Building combo box

hypnoshock:
# What

In the building contruction panel replacing the standard `<select>` drop down list for Adobe Aria's `ComboBox`

image
image

# Why

After a lot of buildings are deployed the list can become unweildy so it's good to be able to search for buildings

# How

I have added a new dependency of `"react-aria-components": "^1.1.1"` and for these to work I have had to update our version of next from `13.2.4` to `13.5.6`. Updating next to a later version appears to be working o.k but I haven't dug into the release notes to see if there is anything that could cause trouble so that might be something that needs to be done before a merge

# Notes

The combo box is styled differently to how it was before. I've tried to keep the style in keeping with the rest of the design but if it's not to our liking it can be restyled very easily as Adobe's aria components are not opinionated about style.


chrisfarms (commented):
nice, that component slot it quite nicely 👍 

code LGTM, will leave @ldunnplaymint to give it some UI abuse and approve

I can't decide if the option-groups are still desirable or not now that you can easily filter... I guess they do still add some context that would otherwise be missing

ldunnplaymint:
It is all kinds of broken for me

https://github.com/playmint/ds/assets/99176798/59e06d43-9dd6-4ca6-ac51-c255f1dc6275

hypnoshock:
> It is all kinds of broken for me
> 
>  Screen.Recording.2024-04-05.at.11.30.16.mov

Cheers Lewis, I'll look into that. I think I had been testing it with lists that fit on the screen. I'll find a way of making that behave, it's going to take a little bit of css magic to knock it into shape

ldunnplaymint:
> > It is all kinds of broken for me
> > Screen.Recording.2024-04-05.at.11.30.16.mov
> 
> Cheers Lewis, I'll look into that. I think I had been testing it with lists that fit on the screen. I'll find a way of making that behave, it's going to take a little bit of css magic to knock it into shape

I should have said that was just using the croissant map as it's the biggest lot of stuff we have

hypnoshock:
> > > It is all kinds of broken for me
> > > Screen.Recording.2024-04-05.at.11.30.16.mov
> > 
> > 
> > Cheers Lewis, I'll look into that. I think I had been testing it with lists that fit on the screen. I'll find a way of making that behave, it's going to take a little bit of css magic to knock it into shape
> 
> I should have said that was just using the croissant map as it's the biggest lot of stuff we have

image

Added scrollbar when the list extends the height of the box. Give that go

ldunnplaymint (approved):
Looks good my G **radical hands**

PR #1249, opened 2024-04-05
Removal of scouting rule

### LINKED ISSUE #1247: do something about ScoutRule
scouting still exists.

scouting could potentially be abused to ruin someone's island at runtime.

we should either remove it, disable it, or think through how it's impact can be mitigated

hypnoshock:
# What

Removal of scouting rule and all associated unit tests

# Why

Scouting doesn't feature in Downstream anymore however the rule left an exploit open where anyone could dispatch a `MOBILE_UNIT_SCOUT` action and spawn tiles on someone's map therefore ruining its shape / cheating

# Notes

In the process of removing scouting I have also removed the perlin noise implementation is we are no longer generating goo values procedually but instead just setting them all to maximum values. The setting of goo values on a tile has been moved to the cheats rule where `DEV_SPAWN_TILE` is implemented.

I have tested on the croissant map that tiles have goo values and I have managed to extract goo without problems

fixes #1247


ldunnplaymint (approved):

Steedie (approved):
Everything seems to be working 👍

PR #1250, opened 2024-04-05
removes "not valid cos..." console log message

Steedie:
## What
Remove the "not valid cos" `console.log` message
image

## Why
It often spams the console and gets in the way when trying to debug things

## Notes
Or we could store the last `reason` to make sure the same message doesn't appear twice in a row?
eg:
```tsx
let lastReason;
...
if (!valid) {
    if (reason != lastReason){
        console.log('not valid cos', reason);
        lastReason = reason;
    }       
}

```

### DIFF
--- a/frontend/src/plugins/action-context-panel/index.tsx
+++ b/frontend/src/plugins/action-context-panel/index.tsx
@@ -770,10 +770,6 @@ const Combat: FunctionComponent<CombatProps> = ({
     const attackTile = path.slice(-1).find(() => true);
 
     const canAttack = mobileUnit && player && valid && attackTile;
-    if (!valid) {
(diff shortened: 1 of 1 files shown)


ldunnplaymint:
I have no problem with removing console spam! It reads like that message will fire if the user is in the attack intent and then selects a tile without a building? I wonder what the motivation behind printing to logs was in that situation, is there anything we might end up missing by killing it outright?

ldunnplaymint (approved):
Approving because it works and that log annoys me. I'd wait for a Farms or someone to make sure there's no reason to keep it there!

chrisfarms:
it was just debug logging that has stuck around from when pathfinding was implemented ... safe to kill

PR #1255, opened 2024-04-08
Limit NAME/DESCRIBE owned entity max characters & annotate -> setData

### LINKED ISSUE #1252: limit NAME_OWNED_ENTITY and DESCRIBE_OWNED_ENTITY
* [x] use setData for storing the name and description data
* [x] update all the graphql queries to fetch from data rather than annotations
* [x] cap the name length to 32chars
* [x] cap the desc lenght to 140ish chars

Steedie:
## What
1. Limited max characters in a name to 32 & description to 140
2. Changed from setting `annotate` to `setData` in `NamingRule.sol`
3. Adjusted code that shows unit name

## Why
1. So that someone can't use their world names and descriptions as a public database!
2. Because farms needs to be able to read name and description data on-chain which `annotate` doesn't currently allow
3. Because `data` is stored as `bytes32` instead of a `string`

## Extra
- Since we removed player naming a while ago, I've removed it from the graphql and implementations of player name in the frontend
- Added alert when mobile unit name is too short (to keep it consistent with behavior when name is too long)
- Mobile unit name prompt now defaults to the current mobile unit name

Resolves #1252

### DIFF
--- a/contracts/src/rules/NamingRule.sol
+++ b/contracts/src/rules/NamingRule.sol
@@ -29,18 +29,22 @@ contract NamingRule is Rule {
     }
 
     function _changeEntityName(State state, bytes24 player, bytes24 entity, string memory name) private {
+        require(bytes(name).length <= 32, "Name exceeds 32 characters");
+
         bytes24 existingOwner = state.getOwner(entity);
         if (existingOwner != 0x0 && existingOwner != player) {
             revert("EntityNotOwnedByPlayer");
         }
-        state.annotate(entity, "name", name);
+        state.setData(entity, "name", bytes32(bytes(name)));
     }
 
     function _changeEntityDescription(State state, bytes24 player, bytes24 entity, string memory desc) private {
+        require(bytes(desc).length <= 140, "Description exceeds 140 characters");
+
         bytes24 existingOwner = state.getOwner(entity);
         if (existingOwner != 0x0 && existingOwner != player) {
             revert("EntityNotOwnedByPlayer");
         }
-        state.annotate(entity, "description", desc);
+        state.setData(entity, "description", bytes32(bytes(desc)));
     }
 }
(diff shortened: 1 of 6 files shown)


ldunnplaymint (approved):

frontend/src/helpers/index.ts
```diff
@@ -15,7 +17,22 @@ export const formatNameOrId = (node?: MaybeNamed, idPrefix: string = ''): string
...
+        // convert the bytes32 (data) value to a Uint8Array
+        const bytes = new Uint8Array(length);
+        for (let i = 0; i < length; i++) {
+            bytes[i] = parseInt(node.name.value.slice(i * 2, i * 2 + 2), 16);
+        }
+
+        // decode the Uint8Array as a string and trim trailing null characters
+        const name = decoder.decode(bytes);
+        const trimmedName = name.replace(/\0/g, '');
+        return trimmedName;
```
chrisfarms, line 32: I think there's an ethers helper for this... we already have ethers as a dep so `import {ethers} from 'ethers'` and something like:

```suggestion
            return ethers.decodeBytes32String(node.name.value);
```

contracts/src/rules/NamingRule.sol
```diff
@@ -29,18 +29,22 @@ contract NamingRule is Rule {
...
         bytes24 existingOwner = state.getOwner(entity);
         if (existingOwner != 0x0 && existingOwner != player) {
             revert("EntityNotOwnedByPlayer");
         }
-        state.annotate(entity, "name", name);
+        state.setData(entity, "name", bytes32(bytes(name)));
     }
 
     function _changeEntityDescription(State state, bytes24 player, bytes24 entity, string memory desc) private {
+        require(bytes(desc).length <= 140, "Description exceeds 140 characters");
```
chrisfarms, line 42: hmmm the 140 is a bit of a lie since we truncate to 32 bytes 😂  .... I guess description needs to be capped at 32chars too

frontend/src/helpers/index.ts
Steedie: much cleaner, thanks!

ldunnplaymint (approved):

PR #1259, opened 2024-04-09
Modify rules/actions to properly handle zone parameter

JackPlaymint:
This PR seems like a bit of an anti-climax, but after looking through I think this is all that needs changing.
I've added a zone parameter check to the distance and isDirect functions in TileUtils.
Everything else seems to parent off from these functions so it should do the trick.

I return a big ol' distance in the distance function if the zones are different. I suppose this could just return an error, but the big distance should serve the purpose just as well without extra error handling.

and IsDirect returns false if the zones are different.

### DIFF
--- a/contracts/src/utils/TileUtils.sol
+++ b/contracts/src/utils/TileUtils.sol
@@ -2,11 +2,13 @@
 pragma solidity ^0.8.13;
 
 import {State, CompoundKeyEncoder, CompoundKeyDecoder} from "cog/IState.sol";
-import {Schema, Node, Kind, Q, R, S} from "@ds/schema/Schema.sol";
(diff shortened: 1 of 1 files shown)


ldunnplaymint (approved):

hypnoshock (approved):
Looks good to me. I got a bit of a PHP flashback when I saw the $ sign on the constant... 😨 however my PTSD isn't enough to hold up this PR ;-)

chrisfarms (approved):
I can't decide if it shoudl be an error or not ... I guess it depends if you are thinking about the "islands" as being on the same sea or in different universes ... heh ... err on the side of quirky heh

Issue #1258, opened 2024-04-09
allow signing in with deployer key during local dev


chrisfarms:
... make it easier to deploy to zone-1 the default deployer owned zone

Issue #1197, opened 2024-03-25
scope queries to a world/zone


chrisfarms:
* Tiles, Buildings, Bags must belong to a Zone 
* new rule/action to register/claim a Zone (unrestricted for now)
* a Player owns a Zone
* update world.graphql query to only fetch Tiles/Buildings/Bags 
* url based zone selection

PR #1260, opened 2024-04-09
Zonership WIP

chrisfarms:
## what

multi zone/island single deployment

* index page is now a (very ugly) palce to list and mint "zones"
* owning the zone NFT gives you DEV_ powers
* use `-z` to target a zone during `ds apply -n local -z $YOUR_ZONE_ID -R -f .`  
* `/zones/${YOUR_ZONE_ID}` renders the game for that zone
* one unit per player restriction now ... spawning will move unit to the zone

## related

* fixes: #1197 
* fixes: #1258 

## WIP

* [ ] bags are very broken ... working on it
* [ ] many maps/plugins are broken as they don't account for the z-th dimension 
* [ ] combat sessions are broken
* [ ] deployer doesn't own the zoneOwnership contract ... it should


ldunnplaymint:
Building fabricator doesn't work because ```no zone data loaded```

ldunnplaymint:
I want a "back zone select" button

chrisfarms:
> I want a "back zone select" button

paul has a button in the "cog" menu, but I don't think there is any plan to place one more prominently

ldunnplaymint:
Probably not a real problem but if I have two zones open in two tabs, spawn my seeker in zone 1, set it on a move path and then swap to the tab with zone 2 and spawn the unit in, the unit will snap back to zone 1 because of the movement chain I have set off.

chrisfarms:
> if I have two zones open in two tabs, spawn my seeker in zone 1, set it on a move path and then swap to the tab with zone 2 and spawn the unit in, the unit will snap back to zone 1 because of the movement chain I have set off.

I declare this a feature not a bug 😂 ... we can't really stop weird behaviour if you are attempting to control a single unit with multiple clients

we decided to keep the move action unrestricted (which is why it teleports you) since we think it'll be more interesting, at the cost of choas like this

hypnoshock (approved):

PR #1263, opened 2024-04-09
Smarter Applier/Destroyer

### LINKED ISSUE #1207: Smarter biome/tile applying
in #1214 we allow undiscovering tiles.

this lets you apply every tile in the zone to either discovered or undiscovered. but applying every tile is an expensive operation.

we should make the cli apply smarter so that it only applies the Tile if it needs to (skips applying a discovered tile if already discovered etc)

Steedie:
## What
When you run `ds apply`, it now gets tiles and buildings from the current state of the world to ensure these things aren't deployed unnecessarily, thus wasting gas. 
image

`ds destroy`, same behavior as above but the opposite ^
image

## Why
To optimise gas and time to deploy.

## How
- It gets the world state and tile state via graphql query (once)
- Logic for filtering gets handled in `getOpsForManifests`

## Update
I removed the smart skipping of buildings that are already applied as it requires more thought to make sure things aren't being overlooked: bags/items in building, facing direction, different name...

Resolves #1207


Steedie:
Changed base to `zonership` so resolving conflicts for that. Going to have to change a bit of the code I think

Steedie:
Conflicts resolved and updated to comply with zone changes - ready for reviews

ldunnplaymint (approved):

cli/src/utils/applier.ts
```diff
@@ -374,6 +378,20 @@ export const getOpsForManifests = async (
...
+        let shouldSkip = false;
+        for (let i = 0; i < convertedBuildingCoords.length; i++) {
+            const coords = String(convertedBuildingCoords[i]).split(',').map(Number);
+            if (coords[1] == q && coords[2] == r && coords[3] == s){
+                shouldSkip = true;
+                skippedBuildings++;
+                break;
+            }
+        }
+
```
chrisfarms, line 390: what happens when there is _a_ building on the tile, but not the same one that is being applied? ... it looks like it will skip it, but not sure if that's what I would expect to happen.

maybe Building are too complicated to be "smart" about here

ldunnplaymint:
I am still seeing it try and fail to apply building and item kinds. It also seems to reapply building instances and succeed.

My repo steps are:
1. Pull this branch
2. make cli
3. Apply tutorial-room-1 folder
4. Reapply tutorial-room-1 folder

Result:
```dispatch fail: failed commit batch tx: execution reverted: already registered
🕒 retrying: registered item Flag

dispatch fail: failed commit batch tx: execution reverted: already registered
❌ registered item Flag```

```dispatch fail: failed commit batch tx: execution reverted: BuildingAlreadyRegistered
🕒 retrying: registered building Fabricated Building

dispatch fail: failed commit batch tx: execution reverted: BuildingAlreadyRegistered
❌ registered building Fabricated Building```

cli/src/utils/applier.ts
Steedie: Removed skipping of applying buildings that are already in the zone ([941ed45](https://github.com/playmint/ds/pull/1263/commits/941ed451ce9b6c79d799e5b0e3153c78671f97b9))

Steedie:
> I am still seeing it try and fail to apply building and item kinds. It also seems to reapply building instances and succeed.
> 
> My repo steps are:
> 
> 1. Pull this branch
> 2. make cli
> 3. Apply tutorial-room-1 folder
> 4. Reapply tutorial-room-1 folder
> 
> Result:
> 
> ```
> 🕒 retrying: registered item Flag
> 
> dispatch fail: failed commit batch tx: execution reverted: already registered
> ❌ registered item Flag```
> 
> ```dispatch fail: failed commit batch tx: execution reverted: BuildingAlreadyRegistered
> 🕒 retrying: registered building Fabricated Building
> 
> dispatch fail: failed commit batch tx: execution reverted: BuildingAlreadyRegistered
> ❌ registered building Fabricated Building```
> ```

I removed the smart skipping buildings in my last commit. This PR now only addresses:
### applier
- smart skipping tiles that are already in the zone

### destroyer
- smart skipping tiles that are not in the zone
- smart skipping buildings that are not in the zone

I left "kinds" out because I think it still needs to reapply if someone wants to make a change to the plugin code.

In your case, it should be failing because you'd need to use the `DEPLOYER_PRIVATE_KEY` to reapply those.

Issue #1265, opened 2024-04-10
Make quests zone aware


ldunnplaymint:
Currently, quests are attached to units but that means they will be taking their quests across zones. We should make it so that quests will only appear in the zone they are relevant to.

PR #1274, opened 2024-04-11
Zoned quests

chrisfarms:
## what

Creates a `ZonedPlayer` entity, to act as a per-player-per-zone point to attach things like quests to.

Attaches quests to that entity, and updates the player query to fetch the zoned data.

Since quests/tasks did not have globally unique ids, I've added the zone into the id and always hash the names to generate the ids to prevent claiming ids that dont conform.

There was no owner check for registering quest/task ids, so I added that

## related

* fixes: #1265 

## notes

until the auto-quest stuff is in you'll have to manually accept the quest from the cli.

* `make dev` to bring up an empty map
* `ds -v -k 0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff9524d487fb5d57a8 apply -n local -z 1 -R -f contracts/src/maps/tutorial-room-3` to apply tutorial 3 to zone 1
* go spawn your unit in zone 1
* `ds -k 0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff9524d487fb5d57a8 -n local -z 1 dispatch ACCEPT_QUEST '["0xadbb33ce00000001000000000000000065f3c5f275072c3c",0]'` to manually accept the auto quest

### DIFF
--- a/cli/src/utils/applier.ts
+++ b/cli/src/utils/applier.ts
@@ -11,7 +11,14 @@ import { AbiCoder, id as keccak256UTF8, solidityPacked } from 'ethers';
 import fs from 'fs';
 import path from 'path';
 import { z } from 'zod';
-import { ContractSource, ManifestDocument, Slot, FacingDirectionTypes } from '../utils/manifest';
+import {
+    ContractSource,
+    ManifestDocument,
+    Slot,
+    Task,
+    FacingDirectionTypes,
+    TaskKindEnumVals,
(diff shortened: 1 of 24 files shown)


chrisfarms:
should be working now ... @hypnoshock I've removed all the keccacking of task kinds, and switched to enums ... dunno what my original issue was but the refactor fixed it 🤷‍♂️

ldunnplaymint (approved):

PR #1277, opened 2024-04-11
Using `setData` to store game settings. Checking ownership via graph

hypnoshock:
# What

- Instead of setting each game setting variable as an edge, using `setData` instead
- Instead of checking ownership via the NFT contract, checking the zone node owner via the graph

# Why

Checking the ownership via the NFT contract directly was convoluted and as the ownership is already recorded on the graph there's a more direct way to look for the zone's owner.

Have an edge to represent each setting wouldn't scale well if we decide to add more settings also calling the game contract to get at them is more expensive than using the `Schema` library


chrisfarms (approved):
noice

PR #1278, opened 2024-04-11
Restyle of index page

hypnoshock:
# What

A light restyle of the index page until we get a proper design:

image

The welcome text is placeholder so will need either rewriting or removing.

It's not great but looks less 'debug' looking


chrisfarms (approved):
I think you overshot "not ugly" this actually looks good! 

non-blocking comments:

* I think maybe toggle buttons rather than a drop-down would be a better fit for the filter since we have all that available horizontal space
* I like that the whole row is clickable ... but maybe still needs a little call to action button or arrow or onhover or something to hint that clicking the row is gonna take me to the zone

but definitely don't make those changes now, this is great as is to get us over the line 👏 👏 👏

PR #1295, opened 2024-04-13
Lazy version of accepting zone quests

chrisfarms:
## what

cherry-picked some bits from @JackPlaymint's work in #1275 to create a quick-n-dirty version that automatically enrolls any player visiting a zone in any zone quests that they do not have.

this is NOT as good as what the other PR was aiming for where it gives the player some UI and option to accept the available zone quests, and we probably still want that too.... but it's a quick fix so that we can use the quest-map today.

**the way quests are assigned really isn't setup for changing things after they are initially deployed (you can't remove quests from existing players, you can't assign a quest to an existing quest index, etc) so I've just made this try it's best then give up.**

PR also contains some boyscouting fixes spotted along the way

### DIFF
--- a/contracts/src/utils/ItemUtils.sol
+++ b/contracts/src/utils/ItemUtils.sol
@@ -41,24 +41,4 @@ library ItemUtils {
     function IDCard() internal pure returns (bytes24) {
         return Node.Item("ID Card", [uint32(100), uint32(100), uint32(76)], true);
     }
-
-    // register is a helper to declare a new kind of item
-    function register(Game ds, ItemConfig memory cfg) internal returns (bytes24) {
-        Dispatcher dispatcher = ds.getDispatcher();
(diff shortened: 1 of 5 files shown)


Steedie (approved):
LGTM, approved.

I had this error in my console when I built `tutorial-room-3` (`MAP=tutorial-room-3 make dev`): 
```
[services] 4:51PM ERR action-rejected-chain error="failed commit batch tx: execution reverted: Quest already present at given questNum" nonce=58017 service=sequencer
[services] 4:51PM ERR remove-stale-pending nonce=58017 opset=0x33d6c232fe161aecb1c47da991c80458c5c82d214509e6edc82f1e04abb243516674a520d8e6f74ea885865948df4ef0cffae654c22fd460320c7d33603aa4181b service=sequencer
```
I tested the quests and everything seems to be working as this PR intends.

chrisfarms:
the 20x rejections is a bit unexpected 😬 ... i did try to make it bail once it gets in that situation ... hopefully this is only temporary until we can make it an explicit player initiated thing to accept (or not) the zone quests 🤞

Issue #1287, opened 2024-04-12
Combat bag UI does not appear


ldunnplaymint:
Kill something, select the bag while your unit is next to it, note the bag UI does not appear

PR #1297, opened 2024-04-13
fix combat bag ui not showing

chrisfarms:
fixes: #1287

same issue that prevented bag models showing on the map... larger mobileunitids were not getting decoded correctly


Steedie (approved):
Nice. Tested `tutorial-room-2`, killed Mr. Squishy, and looted from the combat rewards UI. Approved.

PR #1299, opened 2024-04-13
fix plugins that look for quests

chrisfarms:
quest-map plugins needed to be updated for the zonified quests as the shape of the state object changed


ldunnplaymint (approved):
Looks good :+1: 

I get my quest if I spawn into a zone already with one, if I apply the quest after spawning and they don't follow me around zones.

PR #1304, opened 2024-04-13
Fixed DvB by removing option to choose team buildings

5p0rt5BEArD:
The world.buildingkinds object has gone and DvB was using this to give players a choice of of the building that would be counted for each team. These are now hardcoded instead.

To make the game flexible again I think the plugins would need access to global state - not sure how hard that is to provide or maybe they already do have access.

### DIFF
--- a/contracts/src/maps/duck-burger/DuckBurgerHQ.js
+++ b/contracts/src/maps/duck-burger/DuckBurgerHQ.js
@@ -5,8 +5,6 @@ const prizeItemId = "0x6a7a67f08c72b94400000001000000010000000000000000"; // gre
 const buildingPrizeBagSlot = 0;
 const buildingPrizeItemSlot = 0;
 const nullBytes24 = `0x${"00".repeat(24)}`;
-const duckBuildingTopId = "04";
-const burgerBuildingTopId = "17";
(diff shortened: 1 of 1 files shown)


ldunnplaymint (changes requested):
If I try to have 2 users join the game via the HQ, the plugin stops diplaying anything bar the name/desc and this error fires in the logs 
```client.js:2 plugin did not return an expected response object: TypeError: Cannot read properties of undefined (reading 'filter')
    at getBuildingKindSelectHtml (blob:http://localhost:3000/dc613402-e1a9-42e5-bb45-3fe73ed861cf:570:18)
    at Module.update (blob:http://localhost:3000/dc613402-e1a9-42e5-bb45-3fe73ed861cf:313:15)
    at PluginContext.update (webpack-internal:///./src/workers/snowsphere.ts:56:62)
    at Object.update (webpack-internal:///./src/workers/snowsphere.ts:236:20)
    at callback (comlink.mjs:116:48)```

ldunnplaymint (approved):
Silly Lewis. I initially tested against main so I understood the issue and then didn't retest against the branch, this does actually work now!

PR #1305, opened 2024-04-13
Support applying named foundry compiled artififacts directly

robinbryce:
This has the benefit of exporting the complexities of compiling solidity entirely to the foundry tool chain. And it offers clean integration with external projects which are already setup to use it.

Does nothing unless the manifest directory contains a foundry.toml

If foundry.toml is present, source.file is interpreted as a contract name. FOUNDRY_OUT is read from env and defaults to 'out'. An attempt is then made to locate the compiled artifact under the resulting directory according to the output file schema for foundry.

This could be improved by actually reading foundry.toml to get the out dir

If the derived file doesn't exist, fall back to the previous solc

Example remappings if your manifest dir is a sibling of a ds clone

```
@ds/=../ds/contracts/src/
cog/=../ds/contracts/lib/cog/contracts/src/
```

allow_paths will be needed in foundry.toml. A simple example:

```
[profile.default]
solc_version = "0.8.13"
verbosity = 4
allow_paths = ["../ds/contracts"]
```

### DIFF
--- a/cli/src/commands/apply.ts
+++ b/cli/src/commands/apply.ts
@@ -7,6 +7,31 @@ import { ContractSource, readManifestsDocumentsSync } from '../utils/manifest';
 import { compilePath } from '../utils/solidity';
 import { getGlobal, getZone } from './get';
 
+/**
+ * if contractName identifies a foundry artifact return the bytecode for
+ * deploying the contract.
+ *
+ * @param contractName - The name of the contract.
+ * @param manifestDir - The directory where the manifest is located.
+ * @param outPath - The path within the manifest directory.
+ * @returns The bytecode of the contract if found, as a hex string, otherwise null.
+ */
+function foundryArtifact(contractName: string, manifestDir: string, outPath: string): string | null {
+    const fileName: string = path.join(manifestDir, outPath, `${contractName}.sol`, `${contractName}.json`);
+    if (!fs.existsSync(fileName)) {
+        return null;
+    }
+
+    const content: string = fs.readFileSync(fileName, 'utf-8');
+    try {
+        const hex = JSON.parse(content)["bytecode"]["object"];
+        return hex.slice(2);
+    } catch (err) {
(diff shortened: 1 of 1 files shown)


robinbryce:
Some warts on this PR
* Only really works with -R. manifestDir is derived from -f filename with a straight dirname call when -R is not set, and this breaks this
* If -f somedir is set, the naive join(manifestDir, outDir, ... thing doesn't work terribly well

ds seems pretty determined to own everything  in the plugin directory, so maybe this style of accomodation just isn't going to work

chrisfarms:
I'm concerned we don't have the bandwidth to support multiple toolchains in ds-cli at the moment. But I'm interested to understand the use case more? Is this so that you can deploy a larger project that contains some downstream-related contracts/modifications? or was it some problem with how ds-cli was compiling your code? or something else.

You don't actually _have_ to use the ds-cli and manifests at all, and it is possible to directly call the game actions and deploy things via a foundry script ... maybe an alternative to supporting multiple toolchains in ds-cli, would be to document how you might get started with a foundry script as a kind of advanced path. would something like that help your use case or am I off base?

robinbryce:
Not off base at all. Getting a handle on the scope of use for ds cli was part why I raised the  pr.

if ds cli is made for small self contained plugins , especially for people that have not been exposed to the solidity eco system, and that are not expected to have any externalities, then yeah, just close this.

if you do think supporting projects that are larger, or just have sol dependencies outside of ds, then maybe the way to go is to remove compile support alltogether and integrate with the two major output formats: hardhat and foundry. It’s actually simple for dad to do it that way but it asssumes ‘not novice’ sol skills

A few pointers on self deploying direct, would be very welcome. Even just knowing that is a supported model is great. Thanks

The spirit of this or was sharing a trick it is defo not production grade!

robinbryce:
closed, it makes sense to do this externaly as @chrisfarms suggests

PR #1311, opened 2024-04-15
Zone aware Auto-Quests

JackPlaymint:
This PR adds a UI panel which shows if there is an unaccepted AutoQuest Quest.
It provides a button which allows the player to accept the first quest for that zone.
It also removes the automatic quest accepting code which Farms implemented (although now I'm wondering if that's better? It's closer to what originally happened with AutoQuests ¯\_(ツ)_/¯ )

### DIFF
--- a/frontend/src/components/panels/quest-panel.tsx
+++ b/frontend/src/components/panels/quest-panel.tsx
@@ -338,6 +338,39 @@ export const QuestItem: FunctionComponent<{
     );
 };
 
+export const AutoQuestItem: FunctionComponent<{
+    id: number;
+    questName: string;
+    player: ConnectedPlayer;
+}> = ({ id, questName, player }) => {
+    const onAcceptClick = () => {
+        player
+            .dispatch({
+                name: 'ACCEPT_QUEST',
+                args: [id, 0],
+            })
+            .catch((e) => {
+                console.error('Failed to accept quest', questName, e);
+            });
+    };
+
+    return (
+        <StyledQuestItem expanded={true}>
+            <>
+                <div className="header">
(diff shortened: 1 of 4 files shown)


ldunnplaymint:
The accept quest button doesn't go away after I have accepted one

ldunnplaymint (approved):
Looks good!

frontend/src/components/panels/quest-panel.tsx
```diff
@@ -393,6 +426,17 @@ export const QuestPanel: FunctionComponent<QuestPanelProps> = ({
...
+                (autoquest) =>
+                    !acceptedQuests.find((quest) => quest.node.id === autoquest.id) && (
+                        <AutoQuestItem
+                            key={autoquest.id}
+                            id={autoquest.id || ''}
+                            questName={autoquest.name?.value || ''}
+                            player={player}
+                        />
+                    )
+            )}
```
chrisfarms, line 439: non-blocking: this is a little hard to read... I think this might be easier to read if it filtered the autoquests list, rather than conditionally returning either the AutoQuestItem or undefined .... something like:

```suggestion
            {zone?.autoquests.filter((quest) => acceptedQuests.some((quest) => quest.node.id === autoquest.id)).map((autoquest) => (
                        <AutoQuestItem
                            key={autoquest.id}
                            id={autoquest.id || ''}
                            questName={autoquest.name?.value || ''}
                            player={player}
                        />
                    )
            )}
```

chrisfarms (approved):

ldunnplaymint (approved):

PR #1316, opened 2024-04-22
Allow Buildings to destroy themselves

hypnoshock:
# What

Currently only the owner of the zone can destroy buildings. This PR allows the building kind implementation contract to destroy any instances of its building kind.

# Why

There are games that need to be able to destroy things either as part of the gameplay or as part of a game reset process.
(The selfish reason is that the Turf Wars game would use this)


chrisfarms:
does ZoneKind superceed this? ... I think it's easier to reason about the zone contract handling this stuff, than allowing a buildingkind to globally mess with all instances of it's buildings accross all zones ... but maybe this still has value?

PR #1320, opened 2024-04-22
Fix for ds-cli dependency traversal problem

hypnoshock:
# What

I have added to the `publish` target in the `Makefile` to:
- copy the `dist` folder to `/tmp/ds-cli`
- use jq to remove dependency fields on `package.json` and write the output to `/tmp/ds-cli`
- publish the package from the `/tmp/ds-cli` folder

I have removed the `prepack` script from the `package.json` as npm was trying to build the temporary package again which would fail due to the relative paths to the contracts being incorrect

# Why

When installing `ds-cli` via npm outside of our repo, npm would throw an error when it tried to traverse the `@downstream/core` dependency which is a symlink into a workspace which only exists if running from our main ds repo

# Notes

I have tested this by publishing version 0.0.31
Installing ds-cli as a global package

I checked the `package.json` that's published at https://registry.npmjs.org/@playmint%2fds-cli
I can see that it has published the version without the deps

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ cli: node_modules core/dist/core.js
 	(cd cli && $(NPM) run build && $(NPM) install -g --force .)
 
 publish: cli
-	(cd cli && $(NPM) version patch && $(NPM) publish)
+	(cd cli && $(NPM) version patch && rm -rf /tmp/ds-cli && mkdir -p /tmp/ds-cli && cp -r dist /tmp/ds-cli/dist && jq 'del(.dependencies, .devDependencies)' package.json > /tmp/ds-cli/package.json && cd /tmp/ds-cli && $(NPM) publish)
 
 release: contracts node_modules cli
 	./scripts/release.mjs -i --max-connections 10
--- a/cli/package-lock.json
+++ b/cli/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@playmint/ds-cli",
-  "version": "0.0.29",
+  "version": "0.0.31",
   "lockfileVersion": 2,
   "requires": true,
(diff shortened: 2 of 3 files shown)


ldunnplaymint:
Am I misunderstanding this? 

If I run `npm uninstall @playmint/ds-cli` and then run `npm install -g @playmint/ds-cli` I see the following error:
```npm ERR! Cannot set properties of null (setting 'dev')

npm ERR! A complete log of this run can be found in: /Users/lewisdunn/.npm/_logs/2024-04-22T14_20_13_055Z-debug-0.log```

If i then do `ds version` I can confirm I am on 0.0.29

hypnoshock:
> Am I misunderstanding this?
> 
> If I run `npm uninstall @playmint/ds-cli` and then run `npm install -g @playmint/ds-cli` I see the following error:
> 
> `npm ERR! Cannot set properties of null (setting 'dev')` `npm ERR! A complete log of this run can be found in: /Users/lewisdunn/.npm/_logs/2024-04-22T14_20_13_055Z-debug-0.log`
> 
> If i then do `ds version` I can confirm I am on 0.0.29

I'll look into it, something isn't right then. I've checked the package.json that is in the version published on NPM and it hasn't included any of the dependencies so not sure what is breaking it.

hypnoshock:
It's working now. What was happening was I was manipulating the `package.json` file before it was compressed and then restoring it before it was published which meant that the version of `package.json` within the tarball was correct but the `package.json` registered with npm was the version with all the dependencies!

ldunnplaymint (approved):
No longer seeing the error and I did a quick run at all the commands and I think we are good

PR #1321, opened 2024-04-22
set/get featured zone

### LINKED ISSUE #1314: featured zones
we want to be able to mark one or more zones as "featured" to promote them on the front page.

we want this data to be stored on-chain so that it can be read back easily by analytics from graphql and by watching the transactions

for example this could be an ownerOnly function on the game contract that sets a "featured" edge pointing to a zone.

required for: #1203

Steedie:
## What
- Adds the ability to set a zone as `isFeatured` (true/false)
- Adds "Featured Zones" to the filter drop down
- Adds a "⭐" emoji next to featured zones

image

## Why
We want to have a "Featured" section on the page that shows all the zones.

## Change zone isFeatured on a running local build
An easy way to test this feature is to run this command:
```bash
cast send --rpc-url http://localhost:8545 --private-key 0x6335c92c05660f35b36148bbfb2105a68dd40275ebf16eff9524d487fb5d57a8 0xF8311e28c658003929A7c1218fb8E44cE7A814DE "setZoneIsFeatured(uint64,bool)" 1 true
```

This example sets zone 1 `isFeatured` to `true`.

## Notes
Setting a zone as featured is currently only possible to do by zone owners:
```solidity
    function setZoneIsFeatured(uint64 zoneId, bool isFeatured) public ownerOnly {
        state.setZoneIsFeatured(zoneId, isFeatured);
    }
```

Assuming this should be changed to something like `playmintOnly`, I'm wondering what the best way would be to implement that.

Resolves #1314

### DIFF
--- a/contracts/src/Downstream.sol
+++ b/contracts/src/Downstream.sol
@@ -117,6 +117,7 @@ contract DownstreamGame is BaseGame {
         state.registerEdgeType(Rel.ID.selector, "ID", WeightKind.UINT64);
         state.registerEdgeType(Rel.HasBlockNum.selector, "HasBlockNum", WeightKind.UINT64);
         state.registerEdgeType(Rel.Parent.selector, "Parent", WeightKind.UINT64);
+        state.registerEdgeType(Rel.IsFeatured.selector, "IsFeatured", WeightKind.UINT64);
 
         // create a session router
         BaseRouter router = new DownstreamRouter();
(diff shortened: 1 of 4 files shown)


ldunnplaymint:
Looks like it does what it says on the tin. I am not sure if the intention was for the zone owner to be able to set their zone as featured. I think we, as Playmint, want to be able to spotlight different people's zones at different times.

ldunnplaymint (approved):
I didn't understand owner only, my mistake

chrisfarms:
> Looks like it does what it says on the tin. I am not sure if the intention was for the zone owner to be able to set their zone as featured. I think we, as Playmint, want to be able to spotlight different people's zones at different times.

the `onlyOwner` modifier means "deployerOnly" not zoneOwnerOnly ... so think this is fine

chrisfarms (approved):

Issue #1282, opened 2024-04-12
We should drop wallet connect as a signing method


ldunnplaymint:
It is possible to deploy via wallet connect but it is not possible to buy an Island with wallet connect. We should just support metamask in the cli.

Issue #1267, opened 2024-04-10
cache the session key in localstorage


chrisfarms:
it's annoying having to signin when switching zones, we should cache the session key

PR #1327, opened 2024-04-24
cli: allow authenticating via browser

chrisfarms:
## what

allow ds-cli to get authentication by asking the user to login in a browser window.

makes this the default method for authentication

use `--auth walletconnect` for the old behaviour

## why

so that metamask-extension users have a way of using the cli without needing to setup a walletconnect wallet with the same key in it.

although this is not metamask specific... any method supported by the game login flow is now usable by the ds-cli. for example you could use this to login with a burner, but I don't recommend it

fixes: #1267 
fixes: #1282 

## how

the (slightly mental) flow is:

* a ds cli command that requires dispatch is called ie `ds apply -n $NETWORK -z $ZONE`
* a local http server is started at a well-known port $PORT
* a web browser is opened on the network's game URL (ie localhost:3000 for $NETWORK=local, hexwood1.downstream.game for $NETWORK=hexwood1 etc)
* $NETWORK does the login/auth flow as per the normal game
* $NETWORK redirects the user to the temp local server at `localhost:$PORT/auth=$SESSIONDATA`
* the local http server recvs the $SESSIONDATA and uses it to build the player/dispatcher
* the local http server is shutdown
* cli flow is now authenticated and continues
* a message is left in the browser that it safe to close the window and return to cli

## todo

* [x] considering caching the session key to avoid doing the browser flow for every command
* [x] make the local server messages bit prettier
* [x] call it "browser" not "metamask" auth, it's not just metamask
* [x] do an anvil deploy and test against a remote network
* [x] how can we better invalidate the session data so that stale keys are auto ignored?
* [x] confirm this has not broken the `-k` flow
* [x] remove walletconnect flow from cli
* [x] remove walletconnect flow from game?
* [ ] test works on windows
* [x] test works on linux/wsl

## notes for testing

* you will need to build the cli locally `make cli`
* auth flow will not work against any deployment not deployed from this branch
* since session data is cached, you should clear localstorage in the browser before using any fresh deploy (ie clear localhost:3000 localstorage after a `make dev`, or clear hexwoodX.downstream.game storage after a fresh hexwood deploy)

### DIFF
--- a/chart/templates/app.yaml
+++ b/chart/templates/app.yaml
@@ -13,7 +13,7 @@ networkName: "Garnet"
 networkID: "17069"
 wallets:
   metamask: true
-  walletconnect: true
+  walletconnect: false
   burner: false
 {{ end }}
 zonesAddress: {{ $.Values.frontend.zonesAddress | quote }}
--- a/cli/package.json
+++ b/cli/package.json
@@ -39,6 +39,7 @@
         "cross-fetch": "^4.0.0",
         "fastify": "^4.26.2",
         "glob": "10.3.3",
+        "http-errors": "^2.0.0",
         "lokijs": "^1.5.12",
         "qrcode-terminal": "^0.12.0",
         "solc": "0.8.19",
--- a/cli/package.json
+++ b/cli/package.json
@@ -37,6 +37,7 @@
         "chalk": "^5.3.0",
         "cli-table3": "^0.6.3",
         "cross-fetch": "^4.0.0",
+        "fastify": "^4.26.2",
         "glob": "10.3.3",
         "lokijs": "^1.5.12",
         "qrcode-terminal": "^0.12.0",
--- a/cli/src/main.ts
+++ b/cli/src/main.ts
@@ -47,6 +47,14 @@ yargs
         choices: networks.map((n) => n.name),
         type: 'string',
     })
+    .option('wallet', {
+        alias: 'w',
+        demandOption: true,
+        default: 'metamask',
+        describe: 'wallet authentication method',
+        choices: ['metamask', 'walletconnect', 'private-key'],
+        type: 'string',
+    })
     .option('ws-endpoint', {
         describe: 'override websocket query endpoint',
         type: 'string',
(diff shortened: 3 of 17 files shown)


ldunnplaymint:
Add the active wallet to the browser success message

ldunnplaymint:
If you are not logged into metamask when the cli asks for permission it will just fail. It doesn't wait for you to go through the login/connect flow.

chrisfarms:
ok @ldunnplaymint I have addressed you comments I think, and I believe I have resolved the issue with stale session keys... it should automatically clear useless stored sessions on it's own now, so if you can check things are still good I think this is ready for proper review

last major unchecked things are that I have not tested any of this on Windows ... linux is ok (so long as it's a distro that contains xdg-open which tbf is most of the sane ones).

ldunnplaymint:
I have kicked off a new hexwood14, will ping Paul when it is done.

I am just getting to checking this locally now :+1:

ldunnplaymint:
If I am logged out, apply via cli, go to the cli login flow, reject signing, I am stuck in a forever spinner "Waiting for metamask..."

Closing that popup does nothing (Expected the cli to get sad at some point and fail) and then the connect button no longer functions.

Maybe this is as simple as adding a timeout to the cli?

ldunnplaymint:
We also get depreciation warnings during the "Waiting for metamask" step in the cli:

```(node:35127) [FSTDEP011] DeprecationWarning: Variadic listen method is deprecated. Please use ".listen(optionsObject)" instead. The variadic signature will be removed in `fastify@5`.
(Use `node --trace-deprecation ...` to show where the warning was created)```

chrisfarms:
> We also get depreciation warnings during the "Waiting for metamask" step in the cli:

fixed

chrisfarms:
> If I am logged out, apply via cli, go to the cli login flow, reject signing, I am stuck in a forever spinner "Waiting for metamask..."

this is the case during the normal game login too isn't it? ... I don't think there is a quick fix for this

I don't think the cli should timeout here, since it's waiting for a response, and if you took longer than the timeout to respond you would get a confusing unable to connect message in the browser

Issue #1323, opened 2024-04-22
Add billboard to Hexcraft


PlaymintPaul:
A bit like the DvB one
![image](https://github.com/playmint/ds/assets/105200899/e94f358d-cd20-4944-b0ff-5ea2330a30ce)

Issue #1322, opened 2024-04-22
Add billboard to DvB map


PlaymintPaul:
A bit like this. Wording can be improved a bit
![image](https://github.com/playmint/ds/assets/105200899/71c9c204-78b6-4571-98a9-fdfc75f1e063)

Issue #1318, opened 2024-04-22
Hexcraft map could have the area to the South removed


PlaymintPaul:
Feels... pointless

Issue #1317, opened 2024-04-22
Hexcraft "Start button" does nothing


PlaymintPaul:
Its not enforced, so maybe this building would work better if it gave players a team and nothing else?

E.g.
- Join a Team
- Leave the Game

PR #1331, opened 2024-04-24
Map understandability

PlaymintPaul:
Added How To Play Billboards
Tidied up Hexcraft map
Removed Hexcraft Start Game and made "Reset" clearer

#1323 #1322 #1318 #1317


PlaymintPaul:
> do these bits need copying into the croissant folder?

Not sure. Croissant is a bit of a dead map now. If we ever want to resurrect it we'll also need to take out Tonk. I think it's fine to leave as is at the moment

ldunnplaymint:
I've taken tonk out :+1:

PlaymintPaul:
Ok. Well I'll move things over then. That can be done in another PR

ldunnplaymint (approved):

PR #1332, opened 2024-04-24
Only get src for plugins we care about

JackPlaymint:


core/src/plugins.graphql
```diff
@@ -34,3 +29,20 @@ query GetAvailablePlugins($gameID: ID!) {
         }
     }
 }
+
+query GetPluginSrc($gameID: ID!, $pluginIDs: [String!]!) {
+    game(id: $gameID) {
+        id
+        state(simulated: true) {
+            id
+            plugins: nodes(match: {ids: $pluginIDs}) {
```
chrisfarms, line 38: i forgot that was even possible 😅

chrisfarms (commented):
this looks correct 👍 ... but I now wonder if we have traded one evil (fetching an unbounded list) for another evil (making network requests very frequently).

I think this is good to go in, I'm just not ticking it so it doesn't go out with tonight's  deploy as I'd like to try and setup a test case with a large number of plugins to profile it 📈

JackPlaymint:
**Performance test with 1000 plugins:**
Current plugin loading size 6MB
Screenshot 2024-04-29 at 16 16 19

New plugin loading size 461Kb
Screenshot 2024-04-29 at 16 34 28

ldunnplaymint:
Looks good! I'd already done a smoke test on this PR. Do we want it?

chrisfarms (approved):

Issue #1326, opened 2024-04-23
Implement ZoneKind


ldunnplaymint:
thoughts for kick off...

we want a new Zone manifest:

```yaml
---
kind: Zone
spec:
  name: My Cool Zone 
  description: "Does cool stuff"
  image: http://imgur.com/mycoolimage.png 
  contract:
    file: ./CoolZone.sol
  plugin:
    file: ./CoolZone.js
```

the zone owner can apply the zone manifest which will:

* set the name (required)
* set the description (required)
* set the image url (optional)
* deploy and register a Zone implementation contract (optional)
    * there can only ever be one Zone implementation per zone
    * the Zone implementation should conform the BaseZone/IZone interface (sketched below)
* deploy and register a Zone plugin js (optional)
    * There can only eveer to one Zone plugin per zone 
    * Zone plugins are ALWAYS always active 

A Zone implementation:

* conforms to an interface to allow hooking into common game actions (ie `onMovedMobileUnit()`, `onConstructedBuilding()`, etc)
* the hooks pass along all the relevent args from the original action(s) so that decent decisions can be made
* reverts in hook should revert the original action... but IT MUST NOT BE POSSIBLE TO PREVENT A UNIT LEAVING THE ZONE! 
* ... think: are there any other things that could brick the player?
* there should be a generic `use` type function exposed (just like BuildingKind) and an equivilent `ZONE_USE(actor, payload)` action that can be dispatched by any actor in the zone (I think this means only mobile units? discuss)
* the aim of the `use` is allow an entrypoint for calling the zone contract that is NOT a hook so that for example a plugin could ask the contract to reset stuff

Zone implementation contracts can call DEV_ cheats...

* All the DEV_ actions for a zone should be callable by both the owner AND the zone implementation

hypnoshock:
> IT MUST NOT BE POSSIBLE TO PREVENT A UNIT LEAVING THE ZONE!

If the hook is on arrive and not on leave it won't do. Whilst looking at this yesterday I removed the onLeave hook from buildings as that could trap units

PR #1335, opened 2024-04-24
Zone kinds

hypnoshock:
# What

Adding Zone kinds which can be setup via a zone manifest yaml. The yaml allows us to
- set name
- set description
- specify an implementation contract
- specify a JS plugin

I have also added a variety of hooks into game actions:

```
onUnitArrive
onCombatStart
onCombatJoin
onCombatLeave
onCombatFinalise
onTransferItem
onContructBuilding
onDestroyBuilding
onAcceptQuest
onCompleteQuest
onExtract
onCraft
```

outside of the hooks the zone contract has a `use` function which works similarly to the building kinds `use` function. This means you can call into the zone contract from a building plugin like this:

```
    const destroy = () => {
        const payload = ds.encodeCall("function destroyBuilding(bytes24)", [
            selectedBuilding.id,
        ]);

        ds.dispatch({
            name: "ZONE_USE",
            args: [mobileUnit.id, payload],
        });
    };
```

There is also a new action to `SET_DATA_ON_ZONE` which works the same way as setting data on buildings

# Why

There are times we want to be able to do something globally for a zone and attaching a contract to the zone is a more elegant way of doing this than cramming functionality into an HQ building. Also certain actions such as destroying and creating buildings/tiles can only be done by the owner of the zone; so by allowing the zone contract to dispatch dev actions allows the zone owner to provide the means for players to do such actions which is useful for game resetting etc.

# Note

I haven't put hooks into importing or exporting items. These hooks might be useful however I don't think they can be used to prevent players from exporting items from a zone as the player's unit's bag can carry stuff out of a zone... (if inventory was zoned then I guess we could lock down the ability to import and export items of certain kinds)

fixes: #1326


cli/src/utils/applier.ts
```diff
@@ -35,6 +36,11 @@ const encodePluginID = ({ name }) => {
     return CompoundKeyEncoder.encodeUint160(NodeSelectors.ClientPlugin, id);
 };
 
+const encodeZonePluginID = (zoneId: number) => {
+    const id = Number(BigInt.asUintN(32, BigInt(keccak256UTF8(`zonePlugin/${zoneId}`))));
+    return CompoundKeyEncoder.encodeUint160(NodeSelectors.ClientPlugin, id);
+};
```
chrisfarms, line 42: I'm a little concerned about how you can "squat" these ids by claiming them first, but this is a problem elsewhere in the codebase too, so I don't think it's something we are gonna solve today

contracts/src/maps/zone-test/Zone.sol
```diff
@@ -0,0 +1,49 @@
...
+        Game ds,
+        bytes24 zoneID,
+        bytes24 actor,
+        bytes24[2] memory equipee,
+        uint8[2] memory equipSlot,
+        uint8[2] memory itemSlot,
+        bytes24 toBagId,
+        uint64 qty
+    ) external override {
+        // revert("No transerring in my zone buddy!");
```
chrisfarms: 😆

contracts/src/rules/CheatsRule.sol
```diff
@@ -194,19 +202,45 @@ contract CheatsRule is Rule {
...
+
         bytes24 buildingInstance = Node.Building(z, q, r, s);
+        bytes24 buildingKind = state.getBuildingKind(buildingInstance);
         state.removeBuildingKind(buildingInstance);
         state.removeOwner(buildingInstance);
         state.removeParent(buildingInstance);
         state.removeFixedLocation(buildingInstance);
+
+        // NOTE: There are two places where buildings can be destroyed. CheatsRule and CombatRule
+        // TODO: Maybe this shouldn't call the hook because it's a dev action
```
chrisfarms, line 224: I agree with the comment ... I don't think anything should be able to prevent a DEV_ action ... these are like `sudo` and I don't think anything should stop it

contracts/src/rules/CombatRule.sol
```diff
@@ -226,9 +221,25 @@ contract CombatRule is Rule {
...
         state.setIsFinalised(sessionID, true);
+
+        // Call into the zone kind
+        {
+            (bytes24 attackTileID, /*uint64 combatStartBlock*/ ) = state.getAttackTile(sessionID);
+            bytes24 zone = Node.Zone(state.getTileZone(attackTileID));
+            if (zone != bytes24(0)) {
+                IZoneKind zoneImplementation = IZoneKind(state.getImplementation(zone));
+                if (address(zoneImplementation) != address(0)) {
+                    // TODO: Should we be passing the combatState?
```
chrisfarms, line 232: probably not, it feels like it's maybe internal state that could change and we wouldn't want people depending on it

I think the main thing you want is to be able to know which side won, which you can read from the Session without the state blob right?

contracts/src/rules/InventoryRule.sol
```diff
@@ -163,6 +166,17 @@ contract InventoryRule is Rule {
...
         // perform transfer between item slots
         _transferBalance(state, bags[0], itemSlot[0], bags[1], itemSlot[1], qty);
+
+        // Call into the zone kind
+        {
+            bytes24 zone = Node.Zone(state.getTileZone(location));
+            if (zone != bytes24(0)) {
+                IZoneKind zoneImplementation = IZoneKind(state.getImplementation(zone));
+                if (address(zoneImplementation) != address(0)) {
+                    zoneImplementation.onTransferItem(game, zone, actor, equipee, equipSlot, itemSlot, toBagId, qty);
```
chrisfarms, line 176: I can't imagine what choas this will cause ... I'm here for it 😅

chrisfarms (commented):
looking good 👏

core/src/world.graphql
```diff
@@ -191,6 +191,10 @@ fragment ZoneState on Node {
     url: annotation(name: "url") {
         value
     }
+    allData {
+        name
+        value
+    }
```
chrisfarms, line 197: I'm not sure I get why this is needed? think I'm missing something 🤔

hypnoshock: You can set data on the zone same as you can set data on buildings. There's a new actions `SET_DATA_ON_ZONE`
This is how you'd read it back right? Or have I ballsed something up?

chrisfarms: ah right yeah ... it's for plugins 👍

contracts/src/rules/CheatsRule.sol
hypnoshock: Got rid of that hook. Felt wrong having it there... If builders are calling this from their zoneKind contract they'd know if they called it so a hook not really needed uh

contracts/src/rules/CombatRule.sol
hypnoshock: It wouldn't change as it's the resultant state but I was concerned it might be a lot of data. I'll limit it to the side that's won

chrisfarms:
I know you removed the tutorial test zone as it was just a hacky check ... but can we have a zone manifest with a zone implementaiton (can be noop for every hook) for at least one of the maps? ... maybe quest-map? just as a subsitute for any real docs for the moment and as a quick sanity check

ldunnplaymint (approved):
I have tested quest-map with a zoneKind deployed, all tut islands with zoneKinds and done a tour of all our features while a zoneKind is deployed.

chrisfarms (approved):
I created a Zone.sol that reverts on every hook ...and it reverted as expected 

code LGTM, I'm sure we've probably introduced some quirks somewhere, but this was a heroic PR ... less than 24hrs from idea to working 👏 👏 👏 👏  ... opens up alot of interesting things!

PR #1338, opened 2024-04-24
Only owners can destroy their own bags

hypnoshock:
# What

Fixes bug where anyone could destroy anyone elses bag as the owner was only checked if the bag belonged to a tile

### DIFF
--- a/contracts/src/rules/CheatsRule.sol
+++ b/contracts/src/rules/CheatsRule.sol
@@ -184,15 +184,13 @@ contract CheatsRule is Rule {
         uint8 equipSlot,
         bytes24[] memory slotContents
     ) private {
-        if (bytes4(equipee) == Kind.Tile.selector) {
-            (int16 z, int16 q, int16 r, int16 s) = state.getTileCoords(equipee);
-            require(Bounds.isInBounds(q, r, s), "coords out of bounds");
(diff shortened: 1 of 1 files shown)


contracts/src/rules/CheatsRule.sol
```diff
@@ -189,7 +189,10 @@ contract CheatsRule is Rule {
             require(Bounds.isInBounds(q, r, s), "coords out of bounds");
 
             _checkIsOwnerOrZone(state, ctx, z);
+        } else {
+            require(state.getOwner(equipee) == Node.Player(ctx.sender), "owner only");
```
chrisfarms, line 193: maybe this should just be `else revert` ... ie DEV_DESTROY_BAG can ONLY destroy tile bags .... we could make DEV_SPAWN_BAG only allowed to spawn tile bags for sync ... less powerful, but easir to reason about? 🤔

5p0rt5BEArD: looking at DEV_SPAWN_BAG I think its already the case that only tile bags can be spawned

https://github.com/playmint/ds/blob/7f4f756572ca56502a02b4a1e58356fee0d887de/contracts/src/rules/CheatsRule.sol#L86

or am I mis-reading it?

ldunnplaymint (approved):

PR #1350, opened 2024-04-26
Labyrinth reset

### LINKED ISSUE #1284: Rework labyrinth as dev actions are no longer usable by buildings
Labyrinth in-game resetting no longer works since dev_ actions are restricted to zone owner.

We should remove all the code and put a note in the readme that the owner needs to use the cli to reset after the game has been played.

Blocked: https://github.com/playmint/ds/issues/1324

hypnoshock:
# What

Zone contract implementation for labyrinth so that it can be reset from either the Reset Tower or the Core Tower which is reached upon completion

Also I've changed the `WorldUnit` graphQL fragment to use the annotation `name` instead of the data `name` because decoding the name in the plugin is tricky without access to the ethers library. 

# Why

After locking down who can dispatch DEV actions, the Labyrinth wasn't able to be reset directly from its buildings. The new ZoneKind contract has the power to dispatch DEV actions so it's been refactored to use it

# Notes

The password on the reset tower is `reset` for the time being. 

fixes #1284


5p0rt5BEArD:
image
The goblin post office welcomes the id rather than the name

5p0rt5BEArD:
> image The goblin post office welcomes the id rather than the name

In fact I can't complete the goblin post office - get invalid password

hypnoshock:
> image The goblin post office welcomes the id rather than the name

Did you do a fresh build? There's a fix for that. The graphQL query was getting the unit name as a 32 byte encoded hex which wasn't being decoded in the plugin. Seeing as we don't need the 32 byte hex in the frontend and decoding the string from hex is difficult in the plugin I changed it to fetch the annotation instead.

Also if you're seeing the hex value as the name, Labyrinth will be uncompletable for you as it submits that name as the password

ldunnplaymint:
The reset building doesn't seem to work? I have done the maz and tried using the reset but the button seemingly doesn't fire anything off?

hypnoshock:
> The reset building doesn't seem to work? I have done the maz and tried using the reset but the button seemingly doesn't fire anything off?

Anything in the console?
Are you using `reset` as the password?

hypnoshock:
> In fact I can't complete the goblin post office - get invalid password

That's because it's submitting the encoded string as a hex to the contract. this is happening because the frontend you're using hasn't got the unit name fix which would be the case if you're running it on Garnet.

ldunnplaymint:
Ohhhh, ok, it doesn't refill the bags

ldunnplaymint:
1. No feedback on pressing reset button
2. Doesn't seem to refill the bags that users have taken items from

ldunnplaymint (approved):

Issue #1288, opened 2024-04-12
The create zone button price text doesn't populate if the user signs in via metamask


ldunnplaymint:
Feels like a race condition. It does populate through localdevaccount.

chrisfarms:
I think this is when metamask hasn't switched to the correct network.

we are not very agressive about forcing the switch (because it can cause annoying infinite loops) ... but maybe we need a network banner at the top of the page to indicate you need to switch networks, and we should not show the mint button if we detect you are on wrong network

PR #1355, opened 2024-04-29
MetaMask switch network banner

Steedie:
## What
- A banner that appears at the top of the page whenever you're on the wrong MetaMask network.

image

- Stops it from trying to automatically switch network when connected to MetaMask (replaced for this banner)
  - This was causing problems when buying zones because the user wouldn't immediately know if they were on the correct network or not.

## Notes
There are currently a couple of problems I'm trying to resolve:
- the event listener: `window.ethereum.on('chainChanged', async () => {` is getting added repeatedly
- the `chainChanged` event is not always being called, most likely because it's not being set up at the right time

Resovles #1288


ldunnplaymint:
it works 👍

ldunnplaymint (approved):

PR #1356, opened 2024-04-29
Zoner Info: How to deploy a map

### LINKED ISSUE #1346: Should always be clear how to deploy to your zone
Currently the zone number is a bit hidden. You only really need to know this for deployment purposes. Can we add the deploy command for that zone to the top of the Builder Info Pane (top-right on entering)?

Something like this...
LikeThis

Steedie:
## What
Gives the zone owner information on how to deploy a map

image

## How
The way I'm getting the `-n` value is like this:
```tsx
    const networkName = useConfig()?.networkName;
    const network = useConfig()?.networkName === 'hexwoodlocal' ? 'local' : networkName;
```

Resolves #1346


ldunnplaymint (approved):
Works, assuming there is no better way, it's a :+1:

ldunnplaymint (approved):

Issue #1347, opened 2024-04-25
instrumentation for ds-services


chrisfarms:
we need metrics for:

* actions enqueueing 
* action payload bytes
* time take in simulation
* number of ops in simulation
* time taken to commit
* which block act tx got into 
* sequencer nonce
* sequencer balance
* gas estimation at simulation time
* gas estimation at commit time
* gas spend and limit at commit time
* graphql query response time
* graphql query response size
* errors during indexing
* reverts during sim
* reverts during accept

where possible all the amove should have labels for

* account
* session

PR #1361, opened 2024-05-01
redstone release candidate

chrisfarms:
* rejigs for https://github.com/playmint/cog/pull/35 (no websockets, block source changes, simpler polling queries)
* various tweaks to play nicer with the latencies seen on garnet/redstone deployments
* changes how .devstartup works because it's now required to pass contract addrs to cog services (it can't auto detect)
* dropped the MAP variable as it doesn't really make sense in a zones world ... but mostly because I'm lazy and fixing it is not a priority in this PR
* fixes: #1347 
* fixes: #1375

checklist

* [x] merge #1380 
* [x] merge #1381 
* [x] merge the cog PR https://github.com/playmint/cog/pull/35 
* [x] will building-fab bytecode have changed? - don't think so but we forgot before
* [x] is the docker setup still working? ... I think the entrypoint.sh needs updating to pass in the contracts (like .devstartup)
* [x] publish ds-cli 
* [x] check the 1155 marketplace names
* [x] get the nod from @ldunnplaymint


ldunnplaymint (approved):

hypnoshock (approved):
Green tick from me. I've read through it and there wasn't anything glaringly obvious

JackPlaymint (approved):
I can confirm that this is a large amount of code changes

PR #1362, opened 2024-05-01
Shrine buildings

JackPlaymint:
This PR adds new Shrine building tops and bottoms!
They are super cool and cute!
They can be found in the building fabricator and can also be added into your yaml with the following codes:

Bottoms: 15 and 16
Tops: 18 and 19

So 15-18 would give you one completely new building and 16-19 would give you the other, but of course you can mix and match with any of the other tops and bottoms.


ldunnplaymint (approved):
Turbo cheeky Tom (TCT) giving us double trouble

Issue #1340, opened 2024-04-25
Support resets for DVB & Hexcraft


ldunnplaymint:
Use zoneKinds so we can reset DVB

PR #1379, opened 2024-05-03
Duck Burger ZoneKind game reset

Steedie:
## What

- Ability to reset the duck burger arena
  - recreates buildings like the display buildings, countdown, extractors, etc.
  - destroys weak duck & weak burger buildings within the play area (radius of 5 tiles)

https://github.com/playmint/ds/assets/27741109/4e46b436-725f-4396-ad2b-ee27ed3d329e

## Notes
- I've submitted this as a separate PR to the hexcraft reset because I suspect it will work quite differently because the hexcraft arena is much bigger so I can't search in the radius.

## (Half) Resolves
- #1340


chrisfarms (approved):
this is cool 👏 

It's interesting that you decided to lookup the zone implementation and call it from the building via `BUILDING_USE` ... I hadn't considered that option... we do have the `ZONE_USE` action which let's you dispatch directly to the zone contract, but your way is just as legit

Issue #1307, opened 2024-04-14
Extractors aren't available to build on the Istanbul map


PlaymintDavid:
I don't seem to be able to build any buildings.

5p0rt5BEArD:
The building selection box has changed to a text box. Either click the arrow or start typing. Do you see a building list then?

chrisfarms:
is it related to this: https://github.com/playmint/ds/issues/1303 

(ie the list is empty cos "constructable" is ticked and you don't have enough stuff to build anything)

PlaymintDavid:
Ah yes. I see now. I've been able to progress after I've killed some trees to get goo.

PlaymintDavid:
Uhhh. Can we take a look at this again? It's really confusing that you can't build an extractor. I don't think people will figure out that they need to attack some trees to get the necessary resources and there's nothing in the tutorial text to suggest it.

ldunnplaymint:
Currently, the system will only display what is currently buildable using your resources. The reason we went down this route is because it helped cut down the list and remove noise from the list. Since then, we have added the type search so maybe it isn't as much of an issue as it was. We could turn off the "is craftable?" filter but I think the list will still bloat massively once people start deploying things against the game. I suppose this is one for @PlaymintPaul as it might need more thinking about than what I have written.

PlaymintDavid:
I don't have a problem with the listbox. The problem is that the tutorial is asking to make something that I can't make.

ldunnplaymint:
I change the Goo Harvesting quest text to say "You will need to destroy something/use existing extractors to get more goo"

ldunnplaymint:
Paul and I have discussed this and made some changes to fix the issue

Fixed by: https://github.com/playmint/ds/pull/1344

PlaymintPaul:
I've got a better fix in the Single combat session PR. Deletion tool / registration receipt cost 80/80/80 now

PR #1380, opened 2024-05-03
Refactor combat system to have only one session per tile pair and no historical sessions

### LINKED ISSUE #1364: [Contracts] prevent combat sessions sticking around in the graph
old combat sessions always hang around in the zone query.

after 5 hours of (wrongly) investigating issues with combat I had enough combat session to start noticing the affects of all the old sessions in the data (large query size, long loops)

we should probably change the edges so that finalizing actually disconnects the item from the zone query, either by mounting it on tiles or something else.

this would be a contract change, so not ideal.

I believe we needed them to stick around because of session bags ... however we now how the player wallet, so could send rewards directly there and drop the need for reward bags entirely (which would also be a perf/query win)

hypnoshock:
# What

Refactor combat system to only have one session per tile pair. We no longer keep old sessions and rewards now are transferred directly to the player's wallet. The session query has been moved into the tile fragment because when a session is done with, we don't want fetch the empty session nodes

# Why

Because we kept all old sessions, the sessions list under the zone fragment was forever increasing in size which caused performance issues on the frontend due to data bloat

# Problems

The combat task in quests relies on historical sessions being available to know if the player had won a session / knocked down a builting etc. I don't really have a fix for this other than perhaps writing data into a zone kind and hooking into that somehow. Needs some discussion before this is merged. So that combat tasks aren't incompletable I've made it so they auto complete

fixes #1364


contracts/src/rules/CombatRule.sol
```diff
@@ -233,6 +237,60 @@ contract CombatRule is Rule {
...
+    function _makeRewardBag(State state, bytes24 sessionID) private returns (bytes24 rewardBag) {
+        rewardBag = Node.Bag(CompoundKeyDecoder.UINT64(sessionID));
+        state.setEquipSlot(sessionID, 0, rewardBag);
+        return rewardBag;
+    }
+
+    function _destroyRewardBag(State state, bytes24 sessionID) private {
+        // NOTE: The bag had no owner or parent. It would also be empty so I'm not
+        // going to iterate over the slots to clear them
+        state.setEquipSlot(sessionID, 0, bytes24(0));
```
chrisfarms, line 287: can we remove the edge instead of pointing to the null node?

core/src/world.ts
```diff
@@ -37,7 +37,6 @@ export function makeZone(cog: Source<CogServices>, zoneID: string) {
                 ...flatMapBags(res.game.state.zone.mobileUnits),
                 ...flatMapBags(res.game.state.zone.buildings),
                 ...flatMapBags(res.game.state.zone.tiles),
-                ...flatMapBags(res.game.state.zone.sessions),
```
chrisfarms, line 40: 💪

frontend/src/components/map/CombatSession.tsx
```diff
@@ -2,30 +2,26 @@ import { getTileHeight } from '@app/helpers/tile';
...
-                        id={`session-${t.id}`}
-                        height={getTileHeight(t)}
-                        color="red"
-                        style="gradient_outline"
-                        animation="none"
-                        {...getCoords(t)}
-                    />
-                ));
-        }, [tiles, sessions]);
+export const CombatSessions = memo(({ tiles }: { tiles: WorldTileFragment[] }) => {
```
chrisfarms, line 6: i wonder if we should move adding the highlight into the main tiles loop now, save a couple of loops n memos n hooks ... but dont feel strongly that it will be much of a win

contracts/src/rules/CombatRule.sol
```diff
@@ -233,6 +237,60 @@ contract CombatRule is Rule {
...
+    }
+
+    function _destroySession(State state, bytes24 sessionID) internal {
+        state.setParent(sessionID, bytes24(0));
```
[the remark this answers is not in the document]
hypnoshock: done and done

contracts/src/schema/Schema.sol
```diff
@@ -729,6 +717,15 @@ library Schema {
         state.set(Rel.Has.selector, 0, defenceTile, sessionID, startBlock);
     }
 
+    function unsetCombatTiles(State state, bytes24 sessionID, bytes24 attackTile, bytes24 defenceTile) internal {
+        state.set(Rel.Has.selector, uint8(CombatSideKey.ATTACK), sessionID, bytes24(0), 0);
```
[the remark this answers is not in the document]
hypnoshock: I think I get a bit used to setting things to null in solidity land not thinking that in this case, it's in a map which can have its elements deleted. My mistake - nicely caught!

PlaymintPaul:
- Removed the Deletion Duty quest (and references to it).
- Also changed a few numbers to resolve #1307

chrisfarms (approved):

PR #1391, opened 2024-05-08
Package lock files out of date?

ldunnplaymint:
Noticed these are out of date? @chrisfarms Am I doing it wrong?

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,10 +15,11 @@
   "dependencies": {
     "@inquirer/prompts": "^3.0.2",
     "@inquirer/select": "^1.2.7",
+    "@playmint/ds-cli": "^0.0.34",
     "inquirer": "^9.2.10",
     "octokit": "^3.1.0",
     "semver": "^7.5.4",
-    "yargs": "^17.7.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -15,10 +15,11 @@
   "dependencies": {
     "@inquirer/prompts": "^3.0.2",
     "@inquirer/select": "^1.2.7",
+    "@playmint/ds-cli": "^0.0.34",
```
chrisfarms, line 18: this prop occured because you were in the ds dir and ran npm install without the g flag ... so it installed to the project ... we dont want this dep

package.json
```diff
@@ -15,10 +15,11 @@
...
     "inquirer": "^9.2.10",
     "octokit": "^3.1.0",
     "semver": "^7.5.4",
-    "yargs": "^17.7.2",
-    "yaml": "^2.3.1"
+    "yaml": "^2.3.1",
+    "yargs": "^17.7.2"
```
chrisfarms, line 23: this bit looks like a reordering ... harmless but not needed

Issue #1203, opened 2024-03-25
implement frontpage design


chrisfarms:
* blocked until we actually have a design to implement
* wire up the new design to the previously created basic/ugly version created in #1201 
* implement fetching and listing zones, with some basic stats like number of units located there

PR #1395, opened 2024-05-09
Zones list paint job part one

chrisfarms:
## what

bringing the zone list page inline with design we want ... not implemented any of the filter controls, but this is the basics and should be working...

Screenshot 2024-05-10 at 12 39 56

## why

* resolves: #1203 
* resolves: #1387

## divergance from mock ups

* using very simple toggle buttons for "ALL | ACTIVE | YOURS" instead of the complex filter stuff for now, since we don't have enough data for interesting complex filters
* there is no "sort" options ... things are always sorted by "activity" ... where "activity" is either the number of active units or the time of the last active unit
* filtering by "ACTIVE" means "show me only zones with more than 1 tile that someone has entered at some point before, then sort them by "activity" as above


ldunnplaymint:
If I mint a zone, I don't see it immediately as no one is active. I think the filter should switch to "Yours" if you mint a zone.

chrisfarms:
> If I mint a zone, I don't see it immediately as no one is active. I think the filter should switch to "Yours" if you mint a zone.

👍  ... pushed change that does this

I have also made it so that "ACTIVE" always shows you your own zones even if they are not technically active, as this seems less surprizing

ldunnplaymint:
The mint success text is wrong, always was but now way more obvious
Screenshot 2024-05-10 at 13 25 25

chrisfarms:
> The mint success text is wrong...

what do you want it to say?

ldunnplaymint:
> > The mint success text is wrong...
> 
> what do you want it to say?

"Mint successful! Please fing your Zone in the list below!"

ldunnplaymint:
The bottom playmint button does nothing

ldunnplaymint:
> > The bottom playmint button does nothing
> 
> it's not a button, just a logo 🤷

It looks like a button 🤔

ldunnplaymint (approved):

Issue #1400, opened 2024-05-11
ITEM_USE only works when item is in bag-0-slot-0


chrisfarms:
due to a misused variable here: https://github.com/playmint/ds/blob/main/contracts/src/rules/InventoryRule.sol#L89

ITEM_USE can only be called for an item that is in bag-0-slot-0 or bag-1-slot-1 or bag-2-slot-2 🤦

PR #1401, opened 2024-05-11
expose item data to client

chrisfarms:
## what

ItemKind contracts can `SET_DATA_ITEM` just like BuildingKind contracts can `SET_DATA_BUILDING` to let them store data on-chain and read back in the client.... however we were not previously exposing the item data to the plugins so it was useless.

this exposes the item data so that plugins can do interesting stuff with it.

it updates the cocktail hut so that so "sips" are no longer just a client-local thing and are now a globally shared on-chain counter of all sips ever by everyone.

somewhat hilariously though, there's a bug in the implementation of ITEM_USE that means you can only "use" items that are in either bag-0-slot-0 or bag-1-slot-1 😂  ... so that's a bit annoying and I had to make the cocktail plugin ask you to move the cocktail before using it

Screenshot 2024-05-11 at 14 35 01

issue raised for that #1400 but it would be a contract change so not possible to fix quickly

### DIFF
--- a/contracts/src/maps/cocktail-hut/Cocktail.js
+++ b/contracts/src/maps/cocktail-hut/Cocktail.js
@@ -1,20 +1,64 @@
 import ds from 'downstream';
 
-let numSips = 0;
-
 export default async function update({ selected }) {
+    // find the player's unit
     const { mobileUnit } = selected || {};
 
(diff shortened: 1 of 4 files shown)


5p0rt5BEArD (approved):
lgtm

Issue #1402, opened 2024-05-11
Zones above 40 do not allow units to construct buildings


5p0rt5BEArD:
After entering any Zone with an index of 41 or higher, Units are not able to use the build action.

This is because of a bad require on in the BuildingRule that uses z, q, r to confine build location rather than q, r, s

https://github.com/playmint/ds/blob/2e617cb6d3762a52aedd8a1e879d248e66393221/contracts/src/rules/BuildingRule.sol#L95

PR #1403, opened 2024-05-11
disable construct intent for buggy zones

chrisfarms:
there's [a contract bug](https://github.com/playmint/ds/issues/1402) preventing the CONSTRUCT_BUILDING_MOBILE_UNIT action for working as intended when the zone id is 40 or more.

as a temporary measure we are disabling the build intent for the affected zones so that it reduces the potential confusion of either the resulting "out of bounds" error or the silent failing due to these errors only being debug visisble.

### DIFF
--- a/frontend/src/components/views/shell/index.tsx
+++ b/frontend/src/components/views/shell/index.tsx
@@ -35,7 +35,7 @@ import { styles } from './shell.styles';
 // fails for zone ids above 40. As a temporary measure to limit confusion we
 // will disable the construct button on these zones while we consider options
 // see: https://github.com/playmint/ds/issues/1402
-const isBuggyZone = (id: number) => id >= 40;
+const isBuggyZone = (id: number) => id > 40;
 
 export interface ShellProps extends ComponentProps {}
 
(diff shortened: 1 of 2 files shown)


frontend/src/components/views/shell/index.tsx
```diff
@@ -31,6 +31,12 @@ import styled from 'styled-components';
 import { pipe, subscribe } from 'wonka';
 import { styles } from './shell.styles';
 
+// there is a bug that means that the CONSTRUCT_BUILDING_MOBILE_UNIT action
+// fails for zone ids above 40. As a temporary measure to limit confusion we
+// will disable the construct button on these zones while we consider options
+// see: https://github.com/playmint/ds/issues/1402
+const isBuggyZone = (id: number) => id >= 40;
```
5p0rt5BEArD, line 38: should it be >= 41?

chrisfarms: was erring on the safe side of the off-by one ness

5p0rt5BEArD: I'm going to test it

5p0rt5BEArD: >= 41 is correct

I've confirmed the following two statements locally
- The bug is present on zone 41 but not on zone 40
- The button is not displayed for zone 41 but is displayed for zone 40 with `const isBuggyZone = (id: number) => id >= 41;`

5p0rt5BEArD (approved):
👍

PR #1404, opened 2024-05-11
Scrolly wallet panel

chrisfarms:
quick fix to stop wallet inventory taking over the whole screen and making it unusable


Steedie (approved):
Looks good

PR #1407, opened 2024-05-13
ENS Names used in Zones page and Account panel

Steedie:
## What
Uses ENS name if it can be found instead of the wallet address:
### Zones Page
image

### Account Panel
image

## Notes
- I've tested to see that it's working on the account panel
- ~I don't know 100% if it's working on the zone page until it's in main because I can't mint a zone on my MetaMask account locally (I think?)~
  - However, I can get it to log my ENS on the zones page, so it is working.

### DIFF
--- a/frontend/src/pages/index.tsx
+++ b/frontend/src/pages/index.tsx
@@ -211,7 +211,7 @@ const ZoneList = ({
             });
             observer.current?.disconnect();
         };
-    }, [ensNames, zones]);
+    }, [ensNames]);
 
     return (
         <div>
--- a/frontend/src/pages/index.tsx
+++ b/frontend/src/pages/index.tsx
@@ -1,4 +1,5 @@
 import { ethers } from 'ethers';
+import { lookupENSName } from '@app/helpers';
 import {
     DownstreamLogo,
     EmbossedBottomPanel,
@@ -162,6 +163,20 @@ const ZoneList = ({
     unitZoneLimit: number;
     onClickEnter: (id: number) => void;
 }) => {
+    const [ensNames, setEnsNames] = useState({});
+
+    useEffect(() => {
+        zones.forEach(async (zone) => {
+            if (zone.owner?.addr) {
+                const ensName = await lookupENSName(zone.owner.addr);
+                setEnsNames((prevNames) => ({
+                    ...prevNames,
+                    [zone.owner?.addr]: ensName,
+                }));
+            }
+        });
+    }, [zones]);
+
     return (
         <div>
             {zones.map((zone) => (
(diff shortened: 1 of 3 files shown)


hypnoshock:
I think you should be able to test it locally by adding one of the Anvil accounts to metamask and transferring some of the test ETH to your account. That way you'd be able to mint your own zone. Outside of that, another option could be to mint the zone using the dev account and then transfer the zone 721 to your account by adding the zone721 contract to your Metamask. I think the transfer of fake eth is the easier option though

frontend/src/pages/index.tsx
```diff
@@ -162,6 +163,20 @@ const ZoneList = ({
...
+        zones.forEach(async (zone) => {
+            if (zone.owner?.addr) {
+                const ensName = await lookupENSName(zone.owner.addr);
+                setEnsNames((prevNames) => ({
+                    ...prevNames,
+                    [zone.owner?.addr]: ensName,
+                }));
+            }
+        });
+    }, [zones]);
```
chrisfarms, line 178: i think this hook might get invalidated much more frequently than we would like for the kind of expensive lookup that is happening.

🤔

chrisfarms (commented):
can we confirm how this performs when there are 100s+ of zones with unique addresses?

the page receives data updates multiple times a second, and if each of those updates causes an async external request for hundreds of addrs, it could get hairy.

Steedie:
Could it be better to store it in something like zone.owner.ens when the zone is minted, then update it on their side whenever they connect or join their zone?

chrisfarms:
nah i think you are on the right track with the fetching on demand, we just need to be careful with creating a storm of requests

Steedie:
Since latest commit(s):
- implemented address/ens caching (until page refreshed)
- `lookupENSName` can only be used by 5 zone rows at a time
- only calls `lookupENSName` on zone rows that are visible on the screen

Linting still todo

Steedie:
linted and ready

hypnoshock (approved):
I think this solves the problem of scale that Farms pointed out so LGTM. A comment I have which I think is separate from this PR is about the zone list updating every 2 seconds. I'm not sure if Farms fixed this when he did the big CSS overhaul but in the old version (that I was responsible for 😬) it would rerender all the zone rows each time the zone data came in. I guess we'd need to memoize the zone data as it shouldn't be frequently updating

Steedie:
> I think this solves the problem of scale that Farms pointed out so LGTM. A comment I have which I think is separate from this PR is about the zone list updating every 2 seconds. I'm not sure if Farms fixed this when he did the big CSS overhaul but in the old version (that I was responsible for 😬) it would rerender all the zone rows each time the zone data came in. I guess we'd need to memoize the zone data as it shouldn't be frequently updating

ty :)

Ah, this reminds me that I can remove the `zones` dependency. (latest commit: https://github.com/playmint/ds/pull/1407/commits/8d842a93f27054a062239d2ec40db23398189263)

ldunnplaymint (approved):

PR #1409, opened 2024-05-23
Graphics update

JackPlaymint:
This PR adds some modifications to shaders and batching in order to increase performance.

The most significant change is switching off SRP Batcher and "replacing" it with a better GPU instancing setup.
This should reduce draw calls in the majority of situations, but may not be a "one size fits all" sort of improvement.


ldunnplaymint (approved):
Probably needs a david pass but looks good :+1:

PR #1411, opened 2024-05-24
Fix ducks and building rotation

JackPlaymint:
This PR rotates the duck roofs on both the factory buildings and the duck counter buildings so that the face the same direction as the building itself.

It also adjusts how dynamically placed buildings are rotated so that they show off both of their detailed sides

### DIFF
--- a/frontend/src/components/map/Buildings.tsx
+++ b/frontend/src/components/map/Buildings.tsx
@@ -75,7 +75,8 @@ export const Buildings = memo(
                     const coords = getCoords(b.location.tile);
                     const height = getTileHeightFromCoords(coords);
                     const selected = selectedElementID === b.id ? 'outline' : 'none';
-                    const rotation = b.facingDirection == FacingDirectionKind.RIGHT ? -30 : 30;
+                    console.log(b);
+                    const rotation = b.facingDirection && b.facingDirection == FacingDirectionKind.LEFT ? 30 : -30;
                     const tile = tiles.find(({ id }) => id === b.location?.tile.id);
                     const overrideModel = (pluginBuildingProperties || [])
                         .find((prop) => prop.id == b.id && prop.key == 'model')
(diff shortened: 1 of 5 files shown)


frontend/src/components/map/Buildings.tsx
```diff
@@ -75,7 +75,8 @@ export const Buildings = memo(
                     const coords = getCoords(b.location.tile);
                     const height = getTileHeightFromCoords(coords);
                     const selected = selectedElementID === b.id ? 'outline' : 'none';
-                    const rotation = b.facingDirection == FacingDirectionKind.RIGHT ? -30 : 30;
+                    console.log(b);
```
chrisfarms, line 78: stray console.log 👁️

JackPlaymint: Oopsy Poopsy!

map/ProjectSettings/ScriptableBuildPipeline.json
```diff
@@ -0,0 +1,13 @@
...
+    "cacheServerPort": 8126,
+    "threadedArchiving": true,
+    "logCacheMiss": false,
+    "slimWriteResults": true,
+    "maximumCacheSize": 20,
+    "useDetailedBuildLog": false,
+    "useV2Hasher": true,
+    "fileIDHashSeed": 0,
+    "prefabPackedHeaderSize": 2
+}
```
chrisfarms, line 13: did you mean to commit this? ... seems unrelated

JackPlaymint: I did not mean to commit this... I also don't know what it is 🤔. It may be a left over from playing with the rendering. I'll remove it.

JackPlaymint: Ooookay... turns out it's auto-generated and will appear after a build forever, so I guess we need it. It's also been added in the Graphics update PR I made.

chrisfarms (approved):
dunno if that ScriptableBuildPipeline things should be .gitignored or not ... but will leave it up to you

ldunnplaymint (approved):

ldunnplaymint:
Id gitignore it, easier to add later if its a problem rather than remove it from the list

ldunnplaymint (approved):

